PR #1, opened 2023-04-13
feat: initial jenkins charm

yanksyoon:
This PR creates the initial Jenkins server with wizard setup bypass.

Agent connections and other setup are out of scope for this PR. The web can be access through `http://<unit_ip>:8080`

image


.jujuignore
```diff
@@ -0,0 +1,8 @@
+/env
+/image
+.tox
+*.py[cod]
+*.charm
+image-builder
```
arturo-seijas, line 6: What's this `image-builder`?

metadata.yaml
```diff
@@ -0,0 +1,27 @@
+name: jenkins-k8s-operator
+display-name: Jenkins K8s
+summary: Jenkins Continuous Integration Server
+maintainers:
+  - IS SRE <launchpad.net/~canonical-is-devops>
```
arturo-seijas, line 5: s/IS SRE/IS DevOps

metadata.yaml
```diff
@@ -0,0 +1,27 @@
...
+description: >
+  Jenkins is an extendable open source continuous integration server that
+  monitors executions of repeated jobs. The focus of Jenkins is the
+  building/testing of software project continuously, and monitoring executions
+  of externally-run jobs. More information at http://jenkins-ci.org/.
+
+  This charm provides the Jenkins server service, and when paired with the
+  jenkins agent provides an easy way to deploy Jenkins.
+docs: ""
```
arturo-seijas, line 14: Please, add `source: https://github.com/canonical/jenkins-k8s-operator`

metadata.yaml
```diff
@@ -0,0 +1,27 @@
+name: jenkins-k8s-operator
+display-name: Jenkins K8s
+summary: Jenkins Continuous Integration Server
+maintainers:
+  - IS SRE <launchpad.net/~canonical-is-devops>
+description: >
```
arturo-seijas, line 6: The description should match the format and contents described in https://discourse.charmhub.io/t/changes-to-charmhub-io-pages/9366

pyproject.toml
```diff
@@ -0,0 +1,55 @@
+# Copyright 2022 Canonical Ltd.
```
arturo-seijas, line 1: s/2022/2023

src/charm.py
```diff
@@ -0,0 +1,151 @@
...
+from jenkinsapi.jenkins import Jenkins
+from ops.charm import CharmBase, PebbleReadyEvent
+from ops.main import main
+from ops.model import ActiveStatus, BlockedStatus, Container, MaintenanceStatus
+from ops.pebble import Layer
+
+from path import INITIAL_PASSWORD, JENKINS_HOME, LAST_EXEC, UPDATE_VERSION
+from types_ import Credentials
+
+if TYPE_CHECKING:
```
arturo-seijas, line 22: Why this conditional import?

src/charm.py
```diff
@@ -0,0 +1,151 @@
...
+        }
+        return Layer(layer)
+
+    def _is_jenkins_ready(self) -> bool:
+        """Check if Jenkins webserver is ready.
+
+        Returns:
+            True if Jenkins server is online. False otherwise.
+        """
+        try:
```
arturo-seijas, line 72: How about `return requests.get("http://localhost:8080/login", timeout=10).ok` instead?

src/charm.py
```diff
@@ -0,0 +1,151 @@
...
+        return Credentials(username=user, password=str(password))
+
+    def _unlock_jenkins(self, container: Container) -> None:
+        """Write to executed version and updated version file to bypass Jenkins setup wizard.
+
+        Args:
+            container: The Jenkins container.
+        """
+        credentials = self._get_admin_credentials(container)
+        client = Jenkins("http://localhost:8080", credentials.username, credentials.password)
```
arturo-seijas, line 115: Can you move `http://localhost:8080` to a constant and replace all occurrences?

src/path.py
```diff
@@ -0,0 +1,14 @@
+# Copyright 2023 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+"""Paths used by Jenkins."""
+
+from pathlib import Path
+
+JENKINS_HOME = Path("/var/lib/jenkins")
```
arturo-seijas, line 8: Can we move these to the charm.py file for cohesion?

tests/conftest.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+def pytest_addoption(parser: pytest.Parser):
+    """Parse additional pytest options.
+
+    Args:
+        parser: pytest command line parser.
+    """
+    parser.addoption("--jenkins-image", action="store", default="")
+    parser.addoption("--series", action="store", default="")
```
arturo-seijas, line 16: I don't see the point of parametrizing the series

src/charm.py
yanksyoon: This is strictly a type hint only, so importing the type when not in TYPE_CHECKING will throw an error.

src/path.py
yanksyoon: The [Jenkins charm](https://github.com/jenkinsci/jenkins-charm/blob/master/lib/charms/layer/jenkins/paths.py) has a lot of paths that will be defined later on in the future. Shall I move it to `charm.py` for now and create another module later on? I can foresee that it will need to be split up later on.
Thanks for the comment!

tests/conftest.py
yanksyoon: Indeed, it is not required at this stage. Updating thank you!

arturo-seijas (commented):
Can you ad the license and readme files?

src/charm.py
yanksyoon: This is great! Thank you :D

pyproject.toml
yanksyoon: Updated, thank you!

yanksyoon:
> Can you ad the license and readme files?

The LICENSE already exists in the main branch, I've updated the readme to match the description in the metadata.yaml file. Thank you for the review!

src/charm.py
yanksyoon: Updated, thank you!

metadata.yaml
yanksyoon: Updated thank you!

metadata.yaml
yanksyoon: Woops, my bad, updated thank you!

.jujuignore
yanksyoon: Removed the unused image-builder! Thank you!

src/path.py
arturo-seijas: I'd go with that and see how it evolves. It seems likely that some of the constants will only be used in charm.py, in which case they should be kept there

yanksyoon: Sounds good! Refactored :)

arturo-seijas (dismissed):
LGTM

arturo-seijas (dismissed):
LGTM

arturo-seijas (dismissed):

.jujuignore
```diff
@@ -0,0 +1,7 @@
+/env
+/image
+.tox
+*.py[cod]
+*.charm
+Dockerfile
+Makefile
```
jdkandersson, line 7: Do we need all of these? I don't see a Dockerfile and Makefile?

jenkins_rock/rockcraft.yaml
jdkandersson, line 1: Should there be a cmd or services?

requirements.txt
```diff
@@ -0,0 +1,2 @@
+ops >=2,<3
+jenkinsapi>=0.3
```
jdkandersson, line 2: What is the reason for using this compared to, for example, this: https://pypi.org/project/python-jenkins/?

src/charm.py
```diff
@@ -0,0 +1,156 @@
...
+JENKINS_HOME = Path("/var/lib/jenkins")
+# Path to initial admin password file
+INITIAL_PASSWORD = JENKINS_HOME / Path("secrets/initialAdminPassword")
+# Path to last executed jenkins version file, required to override wizard installation
+LAST_EXEC = JENKINS_HOME / Path("jenkins.install.InstallUtil.lastExecVersion")
+# Path to jenkins version file, required to override wizard installation
+UPDATE_VERSION = JENKINS_HOME / Path("jenkins.install.UpgradeWizard.state")
+
+
+class JenkinsK8SOperatorCharm(CharmBase):
```
jdkandersson, line 37: Should we use the new charm architecture guidance for this charm?

src/charm.py
```diff
@@ -0,0 +1,156 @@
...
+            check_interval: Time in seconds to wait between ready checks.
+
+        Raises:
+            TimeoutError: if Jenkins status check did not pass within the timeout duration.
+        """
+        for _ in range(timeout // check_interval):
+            if self._is_jenkins_ready():
+                break
+            sleep(check_interval)
+        else:
```
jdkandersson, line 97: This is a minor feedback, the for loop ends up calling a final `sleep` if Jenkins never gets ready and then raises which seems to be a bit unexpected. As a operator I would expect a sleep to be followed by a check whether Jenkins is ready or not have the sleep. What do you think?

src/charm.py
```diff
@@ -0,0 +1,156 @@
...
+            raise TimeoutError("Timed out waiting for Jenkins to become ready.")
+
+    def _get_admin_credentials(self, container: Container) -> Credentials:
```
jdkandersson, line 100: Should this kind of logic be in a separate module?

jdkandersson: Similar for the other Jenkins specific business logic

jdkandersson (commented):
Haven't reviewed everything, the feedback so far could lead to significant changes so will leave it there for now

requirements.txt
yanksyoon: I initially thought that library was pointing to an archived library https://github.com/ceph/python-jenkins. I just realized that's a fork. Will replace with that library!

src/charm.py
yanksyoon: Updated! Thank you :D

src/charm.py
yanksyoon: Updated, will leave the conversations unresolved until it gets approval :)

arturo-seijas (approved):
LGTM

jenkins_rock/rockcraft.yaml
```diff
@@ -0,0 +1,42 @@
+name: jenkins
```
amandahla, line 1: Since there is a rock, could you add */*.rock to the .gitignore?

amandahla: Ops, forgot to format ```*/*.rock```

yanksyoon: Added, thank you :)

tests/integration/conftest.py
```diff
@@ -0,0 +1,57 @@
...
+def model_fixture(ops_test: OpsTest) -> Model:
+    """The testing model."""
+    assert ops_test.model
+    return ops_test.model
+
+
+@pytest.fixture(scope="module", name="jenkins_image")
+def jenkins_image_fixture(request: FixtureRequest) -> str:
+    """The OCI image for Jenkins charm."""
+    return request.config.getoption("--jenkins-image")
```
weiiwang01, line 25: How about raise an error here and provide some useful error messages here if the Jenkins image is not provided by command line arguments.

src/jenkins_.py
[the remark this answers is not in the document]
yanksyoon: There was a library getting used for Jenkins API, but it has been removed after the refactor (the library threw errors in places it didn't need to). It is likely to be used later on, but I believe the file can be renamed when that happens. Thanks for noticing that!

src/jenkins_.py
```diff
@@ -0,0 +1,84 @@
...
+JENKINS_HOME = Path("/var/lib/jenkins")
+
+
+def _is_jenkins_ready() -> bool:
+    """Check if Jenkins webserver is ready.
+
+    Returns:
+        True if Jenkins server is online. False otherwise.
+    """
+    return requests.get(f"{JENKINS_WEB_URL}/login", timeout=10).ok
```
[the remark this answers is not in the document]
yanksyoon: Indeed, updated thank you!

tests/integration/conftest.py
yanksyoon: Updated thanks!

weiiwang01 (approved):

metadata.yaml
```diff
@@ -0,0 +1,36 @@
+name: jenkins-k8s-operator
```
arturo-seijas, line 1: I miss 
```
assumes:
  - k8s-api
```

yanksyoon: Updated thanks!

arturo-seijas (approved):

jenkins_rock/rockcraft.yaml
yanksyoon: Added, thanks!

arturo-seijas (commented):
Can you add the `issues.yaml` workflow too?

README.md
```diff
@@ -1 +1,38 @@
-# jenkins-k8s-operator
\ No newline at end of file
+# jenkins-k8s-operator
+
+**This is a WIP project.**
```
arturo-seijas, line 3: Should we remove this to avoid the risk or forgetting to remove it?

yanksyoon: Done and done! Thanks for the reviews!

jenkins_rock/rockcraft.yaml
```diff
@@ -0,0 +1,46 @@
...
+summary: Jenkins rock
+description: Jenkins OCI image for the Jenkins charm
+version: "1.0"
+base: ubuntu:22.04
+build-base: ubuntu:22.04
+license: Apache-2.0
+platforms:
+  amd64:
+services:
```
arturo-seijas, line 13: As I understand it, the `services` are overridden by the pebble plan defined in the charm, meaning we can drop the key

yanksyoon: That's correct, however, I've decided to add the services definition here for easier docker spin up w/ pebble to facilitate rockcraft build and load :) Would that be okay with you?
I do understand the drawbacks of having multiple definitions of pebble services though. 
As far as I can tell from the [list of `rockcraft.yaml`s](https://docs.google.com/spreadsheets/d/1i1G230Ch2r2ZpUT-yrYLEOzqwQd2UZN-FbcmfueypqU/edit#gid=0) we have inside the org, there's no standard on whether to add the service definitions or not.
One without: [OIDC](https://github.com/canonical/oidc-authservice-rock/blob/main/rockcraft.yaml)
One with: [kubeflow](https://github.com/canonical/kubeflow-rocks/blob/main/centraldashboard/rockcraft.yaml).

arturo-seijas: No, there's no consensus, at least for now. Pebble will be always injected as entrypoint, but I'm wondering if, when defining the service in the rockcraft.yaml, it would make more sense to `merge` the new layer with this one during the pebble setup within the charm https://juju.is/docs/sdk/interact-with-pebble

arturo-seijas (dismissed):
LGTM :)

weiiwang01 (approved):

src/jenkins.py
```diff
@@ -0,0 +1,87 @@
...
+
+def _is_jenkins_ready() -> bool:
+    """Check if Jenkins webserver is ready.
+
+    Returns:
+        True if Jenkins server is online. False otherwise.
+    """
+    try:
+        return requests.get(f"{JENKINS_WEB_URL}/login", timeout=10).ok
+    except requests.exceptions.ConnectionError:
```
jdkandersson, line 27: Is this the only kind of error that can occur here?

src/jenkins.py
```diff
@@ -0,0 +1,87 @@
...
+    Args:
+        timeout: Time in seconds to wait for jenkins to become ready in 10 second intervals.
+        check_interval: Time in seconds to wait between ready checks.
+
+    Raises:
+        TimeoutError: if Jenkins status check did not pass within the timeout duration.
+    """
+    start_time = datetime.now()
+    min_wait_seconds = timedelta(seconds=timeout)
+    while True:
```
jdkandersson, line 43: Do you think it would be better to put the condition `now - start_time > min_wait_seconds` here and use an `else` clause to raise the `TimeoutError`? This way we reduce the chance of an infinite loop

src/jenkins.py
```diff
@@ -0,0 +1,87 @@
...
+        if _is_jenkins_ready():
+            break
+        now = datetime.now()
+        if now - start_time > min_wait_seconds:
+            raise TimeoutError("Timed out waiting for Jenkins to become ready.")
+        sleep(check_interval)
+
+
+def get_admin_credentials(password_file_contents: str) -> Credentials:
```
jdkandersson, line 52: Should this instead take the `container` as an argument and return the password?

src/jenkins.py
```diff
@@ -0,0 +1,87 @@
...
+        file.
+
+    Returns:
+        The Jenkins admin account credentials.
+    """
+    user = "admin"
+    return Credentials(username=user, password=password_file_contents.strip())
+
+
+def calculate_env(admin_configured: bool) -> dict[str, str]:
```
jdkandersson, line 66: Should this type hint be more explicit with which keys are in the environment in the type hint? E.g., using `TypedDict`?

src/jenkins.py
```diff
@@ -0,0 +1,87 @@
...
+    """Return a dictionary for Jenkins Pebble layer.
+
+    Args:
+        admin_configured: Whether admin user has been configured in Jenkins.
+
+    Returns:
+        The dictionary mapping of environment variables for the Jenkins pebble service layer.
```
jdkandersson, line 73: Should pebble be mentioned here? This function is intended to be an abstract representation of the environment that should be re-usable outside of pebble?

src/jenkins.py
```diff
@@ -0,0 +1,87 @@
...
+    """
+    env = JenkinsEnvironmentMap(JENKINS_HOME=str(JENKINS_HOME))
+    env["ADMIN_CONFIGURED"] = str(admin_configured)
+    # Mypy type doesn't recognize TypedDict to be compatible with dict.
```
jdkandersson, line 77: This is probably fixed by addressing the comment above

src/types_.py
```diff
@@ -0,0 +1,31 @@
...
+    Attrs:
+        username: The Jenkins account username used to log into Jenkins.
+        password: The Jenkins account password used to log into Jenkins.
+    """
+
+    username: str
+    password: str
+
+
+class JenkinsEnvironmentMap(TypedDict, total=False):
```
jdkandersson, line 21: I guess this could be used for `calculate_env`?

jdkandersson (commented):
Have only reviewed source code, some of the comments may require changes that will also impact the tests so will leave it here for now

jenkins_rock/rockcraft.yaml
```diff
@@ -0,0 +1,46 @@
...
+      - ca-certificates-java
+      - curl
+      - libnss3
+      - unzip
+    overlay-packages:
+      - bash
+      - ca-certificates-java
+      - fonts-dejavu-core
+      - libfontconfig1
+      - openjdk-11-jre-headless
```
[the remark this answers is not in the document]
gregory-schiano: I'd rather use the default JVM supported in the ubuntu base used for this rock, so I'd rather use `default-jre-headless`

jenkins_rock/rockcraft.yaml
```diff
@@ -0,0 +1,46 @@
...
+    override-build: |
+      mkdir -p ${CRAFT_PART_INSTALL}/{srv/jenkins/,etc/default/jenkins/}
+      chown 2000:2000 ${CRAFT_PART_INSTALL}/srv/jenkins
+      # Use jenkins war rather than apt install for easier Jenkins version control.
+      cd ${CRAFT_PART_INSTALL}/srv/jenkins
+      curl -sLO https://mirrors.jenkins.io/war/2.403/jenkins.war
```
gregory-schiano, line 43: We hardcode the version we download here... isn't there a more elegant way ?

metadata.yaml
```diff
@@ -0,0 +1,41 @@
...
+  - ops
+subordinate: false
+containers:
+  jenkins:
+    resource: jenkins-image
+resources:
+  jenkins-image:
+    type: oci-image
+    description: OCI image for Jenkins
+    auto-fetch: true
```
gregory-schiano, line 41: This is not needed, we can remove this

src/charm.py
```diff
@@ -0,0 +1,115 @@
...
+        Args:
+            env: Map of Jenkins environment variables.
+
+        Returns:
+            The pebble layer defining Jenkins service layer.
+        """
+        layer: LayerDict = {
+            "summary": "jenkins layer",
+            "description": "pebble config layer for jenkins",
+            "services": {
```
gregory-schiano, line 67: Shouldn't we add a healthcheck in the pebble plan ?
see: https://github.com/canonical/pebble/blob/master/README.md?plain=1#L519

src/charm.py
```diff
@@ -0,0 +1,115 @@
...
+        """
+        container = event.workload
+        if not container or not container.can_connect():
+            event.defer()
+            return
+
+        self.unit.status = MaintenanceStatus("Configuring Jenkins.")
+        container.add_layer(
+            "jenkins", self._get_pebble_layer(calculate_env(admin_configured=False)), combine=True
+        )
```
gregory-schiano, line 93: Could worth it adding a comment that explains why we start the pebble plan here, and again few lines later as it's not obvious

gregory-schiano:
There's an interesting case in this charm were there's some business logic done in the charm.py as it uses pebble to start a first time jenkins to initialise things and then restarts it again.
In my opinion this business logic shouldn't be in charm.py as it's specific to Jenkins, but we're not supposed to manage the pebble plan from a service module.

Isn't it possible to do the first "start" using a pebble exec instead ? Which will allow to do all this business logic from the jenkins module directly

yanksyoon:
> There's an interesting case in this charm were there's some business logic done in the charm.py as it uses pebble to start a first time jenkins to initialise things and then restarts it again. In my opinion this business logic shouldn't be in charm.py as it's specific to Jenkins, but we're not supposed to manage the pebble plan from a service module.
> 
> Isn't it possible to do the first "start" using a pebble exec instead ? Which will allow to do all this business logic from the jenkins module directly

Thanks for the comment @gregory-schiano!

Interesting point regarding the approach using `pebble exec`! I'm not aware of any initial setup without having to start the Jenkins server. In this case, I'm not sure if `pebble exec` would just hang or affect the later service plan.

jenkins_rock/rockcraft.yaml
yanksyoon: I could take the variable out into an environment variable for the build step. Would that be a bit more explicit?

gregory-schiano:
> > There's an interesting case in this charm were there's some business logic done in the charm.py as it uses pebble to start a first time jenkins to initialise things and then restarts it again. In my opinion this business logic shouldn't be in charm.py as it's specific to Jenkins, but we're not supposed to manage the pebble plan from a service module.
> > Isn't it possible to do the first "start" using a pebble exec instead ? Which will allow to do all this business logic from the jenkins module directly
> 
> Thanks for the comment @gregory-schiano!
> 
> Interesting point regarding the approach using `pebble exec`! I'm not aware of any initial setup without having to start the Jenkins server. In this case, I'm not sure if `pebble exec` would just hang or affect the later service plan.

The official Jenkins operator manages it, I'm digging into it to see how that's managed

src/charm.py
```diff
@@ -0,0 +1,115 @@
...
+    """Charm Jenkins."""
+
+    def __init__(self, *args: Any):
+        """Initialize the charm and register event handlers.
+
+        Args:
+            args: Arguments to initialize the char base.
+        """
+        super().__init__(*args)
+        self.framework.observe(self.on.jenkins_pebble_ready, self._on_jenkins_pebble_ready)
```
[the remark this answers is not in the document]
yanksyoon: I don't think there's a state yet, it should be introduced when relationships start getting built :).

gregory-schiano:
> > > There's an interesting case in this charm were there's some business logic done in the charm.py as it uses pebble to start a first time jenkins to initialise things and then restarts it again. In my opinion this business logic shouldn't be in charm.py as it's specific to Jenkins, but we're not supposed to manage the pebble plan from a service module.
> > > Isn't it possible to do the first "start" using a pebble exec instead ? Which will allow to do all this business logic from the jenkins module directly
> > 
> > 
> > Thanks for the comment @gregory-schiano!
> > Interesting point regarding the approach using `pebble exec`! I'm not aware of any initial setup without having to start the Jenkins server. In this case, I'm not sure if `pebble exec` would just hang or affect the later service plan.
> 
> The official Jenkins operator manages it, I'm digging into it to see how that's managed

Couldn't dig too much but as far as I see it works using a kind of init groovy script:
https://github.com/jenkinsci/kubernetes-operator/blob/44a7d2460a97a74ae1dd21e286a342bec805c794/pkg/configuration/base/resources/init_configuration_configmap.go

Probably injected using these hooks:
https://www.jenkins.io/doc/book/managing/groovy-hook-scripts/

jenkins_rock/rockcraft.yaml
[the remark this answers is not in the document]
jdkandersson: that can be ok, using `default-jre-headless` did cause problems with Jenkins in the past where Jenkins dropped support for a version of Java which was the default for the Ubuntu version. I guess one or the other our integration tests would spot it and `default-jre-headless` should not change on a given Ubuntu base

src/charm.py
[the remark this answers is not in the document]
jdkandersson: ok

jenkins_rock/rockcraft.yaml
jdkandersson: Part of the intent is actually to hardcode the Jenins version (major.minor). The spec outlines the reason for it if you are interested

src/charm.py
```diff
@@ -0,0 +1,115 @@
...
+from jenkins import JENKINS_HOME, calculate_env, get_version, wait_jenkins_ready
+
+if TYPE_CHECKING:
+    from ops.pebble import LayerDict  # pragma: no cover
+
+
+logger = logging.getLogger(__name__)
+
+# Path to initial admin password file
+INITIAL_PASSWORD = JENKINS_HOME / Path("secrets/initialAdminPassword")
```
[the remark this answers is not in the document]
yanksyoon: fixed, thank you!

requirements.txt
```diff
@@ -0,0 +1,2 @@
+ops >=2,<3
+requests
```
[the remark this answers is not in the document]
yanksyoon: fixed, thank you!

jenkins_rock/rockcraft.yaml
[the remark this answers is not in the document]
yanksyoon: fixed, thank you!

.jujuignore
```diff
@@ -0,0 +1,3 @@
+.tox
+*.py[cod]
```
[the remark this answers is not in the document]
yanksyoon: removed, thank you!

src/charm.py
```diff
@@ -0,0 +1,115 @@
...
+        super().__init__(*args)
+        self.framework.observe(self.on.jenkins_pebble_ready, self._on_jenkins_pebble_ready)
+
+    def _unlock_jenkins(self, container: Container) -> None:
+        """Write to executed version and updated version file to bypass Jenkins setup wizard.
+
+        Args:
+            container: The Jenkins container.
+        """
+        version = get_version()
```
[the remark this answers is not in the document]
yanksyoon: fixed, thank you!

src/jenkins.py
yanksyoon: fixed, thank you!

src/jenkins.py
yanksyoon: fixed, thank you!

src/jenkins.py
yanksyoon: fixed, thank you!

src/jenkins.py
yanksyoon: updated description thank you!

src/types_.py
yanksyoon: fixed, thank you!

src/charm.py
yanksyoon: added health check, thank you!

jenkins_rock/rockcraft.yaml
yanksyoon: refactored out to be a build environment variable. thank you!

metadata.yaml
yanksyoon: Removed, thanks!

README.md
```diff
@@ -1 +1,36 @@
...
\ No newline at end of file
+# jenkins-k8s-operator
+
+A [Juju](https://juju.is/) [charm](https://juju.is/docs/olm/charmed-operators)
+deploying and managing [Jenkins](https://jenkins.io/) on Kubernetes. Jenkins is an open source
+automation server, providing plugins to support building, deploying and automating any project.
+
+Jenkins is an extendable open source continuous integration server that
+monitors executions of repeated jobs. The focus of Jenkins is the
+building/testing of software project continuously, and monitoring executions
```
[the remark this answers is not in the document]
yanksyoon: updated, thank you!

README.md
```diff
@@ -1 +1,36 @@
-# jenkins-k8s-operator
\ No newline at end of file
+# jenkins-k8s-operator
+
+A [Juju](https://juju.is/) [charm](https://juju.is/docs/olm/charmed-operators)
+deploying and managing [Jenkins](https://jenkins.io/) on Kubernetes. Jenkins is an open source
+automation server, providing plugins to support building, deploying and automating any project.
+
+Jenkins is an extendable open source continuous integration server that
```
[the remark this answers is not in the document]
yanksyoon: updated thank you!

src/jenkins.py
yanksyoon: Updated to include TimeoutError, thank you!

src/charm.py
```diff
@@ -0,0 +1,115 @@
...
+            event.defer()
+            return
+
+        self.unit.status = MaintenanceStatus("Configuring Jenkins.")
+        container.add_layer(
+            "jenkins", self._get_pebble_layer(calculate_env(admin_configured=False)), combine=True
+        )
+        container.replan()
+        try:
+            wait_jenkins_ready()
```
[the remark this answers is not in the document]
yanksyoon: I'll rewrite `unlock_jenkins`  function to be a groovy script execution to match the method suggested by @gregory-schiano, as the official kubernetes operator's method.

src/charm.py
yanksyoon: I'll rewrite `unlock_jenkins`  function to be a groovy script execution to match the method suggested by @gregory-schiano, as the official kubernetes operator's method.

src/jenkins.py
```diff
@@ -39,43 +45,43 @@ def wait_jenkins_ready(timeout: int = 140, check_interval: int = 10) -> None:
...
 
     Args:
-        password_file_contents: The contents inside `$JENKINS_HOME/secrets/initialAdminPassword`
-        file.
+        connectable_container: Connectable container to interact with filesystem.
 
     Returns:
         The Jenkins admin account credentials.
     """
     user = "admin"
```
jdkandersson, line 68: Is there any chance this could be changed by the operator?

tests/integration/conftest.py
```diff
@@ -0,0 +1,59 @@
...
+    """The testing model."""
+    assert ops_test.model
+    return ops_test.model
+
+
+@pytest.fixture(scope="module", name="jenkins_image")
+def jenkins_image_fixture(request: FixtureRequest) -> str:
+    """The OCI image for Jenkins charm."""
+    jenkins_image = request.config.getoption("--jenkins-image")
+    assert jenkins_image, "--jenkins-image argument was not provided."
```
jdkandersson, line 26: Do you think this should tell the developer that the image is required?

jdkandersson: At the moment it is saying "not provided" compared to a nice user message such as "the --jenkins-image argument is required which should contain the name of the OCI image" or something like that?

tests/integration/conftest.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+@pytest_asyncio.fixture(scope="module", name="unit_ip")
+async def unit_ip_fixture(model: Model, application: Application):
+    """Get Jenkins charm unit IP."""
+    status: FullStatus = await model.get_status([application.name])
+    for unit in status.applications[application.name].units.values():
+        assert unit, "Invalid unit status."
+        assert unit.address, "Unit does not have an assigned IP."
+        return str(unit.address)
```
jdkandersson, line 53: Is this supposed to be a yield? Otherwise why is it in a `for` loop please?

tests/unit/conftest.py
```diff
@@ -0,0 +1,149 @@
...
+    """Enable ops test framework harness."""
+    harness = Harness(JenkinsK8SOperatorCharm)
+    yield harness
+    harness.cleanup()
+
+
+@pytest.fixture(scope="function", name="jenkins_version")
+def jenkins_version_fixture():
+    """Jenkins version fixture."""
+    return "2.400"
```
jdkandersson, line 36: I think this is defined twice (also in the rockcraft?) Is there any way we could have the Jenkins version in just 1 place?

tests/unit/conftest.py
```diff
@@ -0,0 +1,149 @@
...
+
+        Args:
+            url: URL placeholder parameter to match original function signature.
+            status_code: Status code of the returned response.
+            kwargs: Other keyword arguments to match function signature.
+
+        Returns:
+            Mocked response.
+        """
+        del url, kwargs
```
jdkandersson, line 54: What is the reason for these `del`?

tests/unit/conftest.py
```diff
@@ -0,0 +1,149 @@
...
+    # if the path is an absolute path starting at root / directory, we must make it relative
+    # path, otherwise the overloaded path append operator (/) doesn't work.
+    initial_password_path = make_relative_to_path(tmp_path, JENKINS_PASSWORD_FILE_PATH)
+    initial_password_path.parent.mkdir(exist_ok=True, parents=True)
+    initial_password_path.write_text(admin_credentials.password, encoding="utf-8", newline="\n")
+    return tmp_path
+
+
+@pytest.fixture(scope="function", name="mocked_container")
+def mocked_container_fixture(
```
jdkandersson, line 81: Is this the same as what we had in the 12-factor charm? Perhaps we can put it as a pattern somewhere in the contributing guide repo?

jdkandersson: And mentioned where it came from here so we don't review the same code multiple times 🙂

tests/unit/conftest.py
```diff
@@ -0,0 +1,149 @@
...
+    ) -> None:
+        """Mocked container push function.
+
+        Args:
+            path: Path to push the files to.
+            source: Content to write to given path.
+            encoding: Encoding of the given source content.
+            kwargs: Other keyword arguments to match function signature.
+        """
+        del kwargs
```
jdkandersson, line 123: If this is to not have pylint complain, do you think a pylint disable would be better?

jdkandersson: Or just have `**_` which shoul make pylint happy

tests/unit/helpers.py
```diff
@@ -0,0 +1,43 @@
...
+# See LICENSE file for licensing details.
+
+"""Helper functions used to unit test Jenkins charm."""
+
+from pathlib import Path
+
+import requests
+
+
+def make_relative_to_path(tmp_path: Path, root_path: Path) -> Path:
```
jdkandersson, line 11: What is the reason for needing this please?

yanksyoon:
@gregory-schiano The operator uses jenkins client to run the groovy script set by the configmap. Source code [here](https://github.com/jenkinsci/kubernetes-operator/blob/master/pkg/groovy/groovy.go#L45). Will update the code.

tests/unit/helpers.py
```diff
@@ -0,0 +1,43 @@
...
+        """Placeholder init function to match function signatures.
+
+        Args:
+            args: Placeholder args.
+            kwargs: Placeholder kwargs.
+
+        Raises:
+            ConnectionError: To mock connection error.
+        """
+        del args, kwargs
```
jdkandersson, line 42: Here as well, either `_` or a pylint disable would be clearer?

tests/unit/test_charm.py
```diff
@@ -0,0 +1,103 @@
...
+
+import jenkins as jenkins_src
+from charm import JenkinsK8SOperatorCharm
+from jenkins import LAST_EXEC, UPDATE_VERSION, unlock_jenkins
+
+from .helpers import make_relative_to_path
+from .types_ import ContainerWithPath
+
+
+def test_unlock_jenkins(
```
jdkandersson, line 26: Should this be in the `test_jenkins.py` file? It seems to test a function from that module?

tests/unit/test_charm.py
```diff
@@ -0,0 +1,103 @@
...
+
+
+@pytest.mark.parametrize(
+    "status_code,expected_status",
+    [
+        pytest.param(503, BlockedStatus, id="jenkins not ready"),
+        pytest.param(200, ActiveStatus, id="jenkins ready"),
+    ],
+)
+def test__on_jenkins_pebble_ready(  # pylint: disable=too-many-arguments
```
jdkandersson, line 77: Should this, and any other disables, have an explanatory comment on them?

tests/unit/test_charm.py
```diff
@@ -0,0 +1,103 @@
...
+    status_code: int,
+    expected_status: StatusBase,
+):
+    """
+    arrange: given a mocked jenkins client and a patched requests instance.
+    act: when the Jenkins pebble ready event is fired.
+    assert: the unit status should show expected status.
+    """
+    # speed up waiting by changing default argument values
+    monkeypatch.setattr(jenkins_src.wait_jenkins_ready, "__defaults__", (1, 1))
```
jdkandersson, line 91: This seems a bit prone to error since the order of the arguments could be changed or new ones added, is there any other way of achieving this? E.g., could this be state?

tests/unit/test_jenkins.py
```diff
@@ -0,0 +1,140 @@
...
+    mocked_get_request: Callable[[str, int, Any, Any], requests.Response],
+):
+    """
+    arrange: given mocked requests that returns a 200 response.
+    act: wait for jenkins to become ready.
+    assert: No exceptions are raised.
+    """
+    monkeypatch.setattr(requests, "get", partial(mocked_get_request, status_code=200))
+
+    wait_jenkins_ready(1, 1)
```
jdkandersson, line 90: No `assert`?

tests/unit/test_jenkins.py
```diff
@@ -0,0 +1,140 @@
...
+    """
+    assert get_admin_credentials(mocked_container) == admin_credentials
+
+
+@pytest.mark.parametrize(
+    "admin_configured, expected_map",
+    [
+        pytest.param(
+            False,
+            {"JENKINS_HOME": str(JENKINS_HOME_PATH), "ADMIN_CONFIGURED": "False"},
```
jdkandersson, line 107: Since only `ADMIN_CONFIGURED` changes, consider making the parameter just that and constructing the rest of the dictionary in the test

src/jenkins.py
```diff
@@ -39,43 +45,43 @@ def wait_jenkins_ready(timeout: int = 140, check_interval: int = 10) -> None:
...
+    password_file_contents = str(connectable_container.pull(JENKINS_PASSWORD_FILE_PATH).read())
```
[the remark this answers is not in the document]
yanksyoon: It returns `bytes | str` :)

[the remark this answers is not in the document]
jdkandersson: If it is `bytes`, we would need to `.decode(encoding="utf-8")`, right?

tests/integration/conftest.py
yanksyoon: It just needs to return any one of the unit IPs. Currently `status.applications[application.name].units` cannot be subscripted (hence I can't just take out units[0]). This was the most straightforward way to get the IPs.

src/types_.py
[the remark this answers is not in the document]
jdkandersson: Ah, `total=False` should generally only be used for actually optional values, not to help with progressively instantiating. You can create `TypedDicts` just like a normal dict and put a type hint on it, e.g.:

```python
env: JenkinsEnvironmentMap = {"<key 1>": "<value 1", ...}
```

tests/integration/conftest.py
jdkandersson: What about using https://docs.python.org/3/library/functions.html#next? That eliminates the need for `for`?

tests/unit/test_jenkins.py
yanksyoon: `wait_jenkins_ready(1, 1)` doesn't return any value (None), the test just makes sure that no exceptions are raised. If `assert wait_jenkins_ready(1,1)` is used, mypy would complain `"wait_jenkins_ready" does not return a value  [func-returns-value]`

src/jenkins.py
yanksyoon: Not at the moment, it cannot be changed.

tests/unit/test_jenkins.py
yanksyoon: fixed, thanks!

tests/unit/test_charm.py
yanksyoon: fixed, thanks!

tests/unit/test_charm.py
yanksyoon: fixed, thanks!

tests/unit/helpers.py
yanksyoon: fixed, thanks!

tests/unit/conftest.py
yanksyoon: fixed, thanks!

tests/unit/conftest.py
yanksyoon: Updated with underscores, thanks!

tests/unit/conftest.py
yanksyoon: Updated to refer from rockcraft yaml, thanks!

tests/integration/conftest.py
```diff
@@ -0,0 +1,59 @@
...
+async def application_fixture(ops_test: OpsTest, model: Model, jenkins_image: str) -> Application:
+    """Build and deploy the charm."""
+    # Build and deploy charm from local source folder
+    charm = await ops_test.build_charm(".")
+    resources = {"jenkins-image": jenkins_image}
+
+    # Deploy the charm and wait for active/idle status
+    application = await model.deploy(charm, resources=resources, series="jammy")
+    await model.wait_for_idle(
+        apps=[application.name], status="active", raise_on_blocked=True, timeout=1000
```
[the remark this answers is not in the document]
yanksyoon: Woops! my bad, it's supposed to be 100. Updated thanks!

tests/integration/conftest.py
yanksyoon: Updated thanks!

src/types_.py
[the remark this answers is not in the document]
yanksyoon: TIL! Thank you for sharing that!

src/jenkins.py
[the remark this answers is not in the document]
yanksyoon: Updated to read as str (by giving encoding parameter). Thank you!

src/jenkins.py
```diff
@@ -39,43 +45,43 @@ def wait_jenkins_ready(timeout: int = 140, check_interval: int = 10) -> None:
...
-    while True:
+    while now - start_time < min_wait_seconds:
         if _is_jenkins_ready():
             break
         now = datetime.now()
-        if now - start_time > min_wait_seconds:
-            raise TimeoutError("Timed out waiting for Jenkins to become ready.")
         sleep(check_interval)
+    else:
+        raise TimeoutError("Timed out waiting for Jenkins to become ready.")
```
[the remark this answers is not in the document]
yanksyoon: Updated, thanks!

tests/unit/test_charm.py
yanksyoon: I'm not quite sure I understand what you mean.

I've tried introducing `is_jenkins_ready` as a Jenkins state
```
from jenkins import is_jenkins_ready

class State:
...
@property
def is_jenkins_ready(self):
    return is_jenkins_ready()
```

However, the wait_for_jenkins would still need to be configured with default `timeout` & `interval` parameters which this patch tries to speed up.

Would there be other ways to speed the `wait_for_jenkins` up?

yanksyoon:
@gregory-schiano After some reading through the `kubernetes-operator` repo for [Jenkins](https://github.com/jenkinsci/kubernetes-operator/blob/44a7d2460a97a74ae1dd21e286a342bec805c794/controllers/jenkins_controller.go#L391), it seems that the wizard is bypassed using `-Djenkins.install.runSetupWizard=false` rather than setting up files after wizard being run. The groovy script referred [here](https://github.com/jenkinsci/kubernetes-operator/blob/44a7d2460a97a74ae1dd21e286a342bec805c794/pkg/configuration/base/resources/init_configuration_configmap.go) (the previous link to the configmap.go file) is a script to setup initial user.
For now, I think the current setup should be working ok. Groovy scripts will have to be added later though for upcoming tasks. I don't mind changing the implementation to be using the same Groovy script as the offical kubernetes operator repo. What are your thoughts?

tests/unit/conftest.py
yanksyoon: Refactored to match 12-factor charm testing method, will add adding to the contributing guide for TODO for tomorrow!

tests/unit/helpers.py
yanksyoon: Removed w/ refactored Harness container!

tests/unit/test_charm.py
jdkandersson: I was more thinking that the timeout and interval could be state, although it would be a bit of a stretch. If it was state it could be changed before calling the function

tests/unit/helpers.py
```diff
@@ -3,41 +3,17 @@
...
-        root_path = Path(str(root_path).replace("/", "", 1))
-    return tmp_path / root_path
-
-
+# There aren't enough public methods with this patch class.
 class ConnectionExceptionPatch:  # pylint: disable=too-few-public-methods
     """Class to raise ConnectionError exception."""
 
-    def __init__(self, *args, **kwargs) -> None:
+    def __init__(self, *_, **__) -> None:
```
jdkandersson, line 13: I think it is fine to use `_` for both?

tests/unit/test_charm.py
yanksyoon: I see. In my opinion, it would be better to keep the state clear of such parameters :)

tests/unit/helpers.py
yanksyoon: image
I'm getting a "duplicate parameter" error when using a `_` for both. :D It would be nice to be able to just keep all unused vars `_` though!

jdkandersson: Thats pylance being wrong 😉

yanksyoon:
```
multiprocessing.pool.RemoteTraceback: 
"""
Traceback (most recent call last):
  File "/home/yanks/Documents/canonical/jenkins-k8s-operator/.tox/lint/lib/python3.10/site-packages/pydocstyle/parser.py", line 427, in parse
    compile(src, filename, 'exec')
  File "/home/yanks/Documents/canonical/jenkins-k8s-operator/tests/unit/test_jenkins.py", line 101
    def __init__(self, *_, **_) -> None:
                             ^
SyntaxError: duplicate argument '_' in function definition
```
Hm... I get a syntax error too when I try to run `tox` after the change. Would a variable name such as `unused` be preferred here? https://peps.python.org/pep-0640/

jdkandersson: I guess `*_args` and `**_kwargs` would also work

jdkandersson: And yes it seems Python doesn't allow duplicate input arguments
Screenshot 2023-05-12 at 12 19 38 pm

yanksyoon: Changed! Thank you :D

src/charm.py
yanksyoon: Added comments explaining the restarts. Thank you!

jdkandersson (approved):

weiiwang01 (approved):

PR #2, opened 2023-05-24
feat: k8s agent charm integration

yanksyoon:
This PR provides integration of the Jenkins server charm into the Jenkins k8s agent charm.

image

Additional code changes include:

-  Import `jenkins` module rather than individual imports, i.e. `from jenkins import ...`, to allow source of function to be clearer (improve readability).
- Renaming of `jenkins` module functions. `wait_jenkins_ready` -> `wait_ready`
- Refactoring `types_.py`, migrate types to relevant modules.
- `get-admin-password` action to facilitate `jenkinsapi.jenkins.Jenkins` client creation for testing.


metadata.yaml
```diff
@@ -38,3 +38,7 @@ resources:
   jenkins-image:
     type: oci-image
     description: OCI image for Jenkins
+requires:
+  agent:
+    interface: jenkins-slave
```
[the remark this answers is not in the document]
yanksyoon: Added woke.yaml since this interface cannot be quite changed yet.

.trivyignore
```diff
@@ -1,6 +1,14 @@
 # Jenkins CVEs
-CVE-2023-20863
 CVE-2016-1000027
+CVE-2023-20863
```
amandahla, line 3: Is not possible to address these issues?

yanksyoon: The issues are with things packaged in the jenkins.war file which is already the latest version. Perhaps these issues will get resolved in later patches!

src/charm.py
```diff
@@ -6,36 +6,57 @@
...
+
         self.framework.observe(self.on.jenkins_pebble_ready, self._on_jenkins_pebble_ready)
+        self.framework.observe(self.on["agent"].relation_joined, self._on_agent_relation_joined)
+        self.framework.observe(self.on.get_admin_password_action, self._on_get_admin_password)
 
-    def _get_pebble_layer(self, jenkins_env: JenkinsEnvironmentMap) -> Layer:
+    @property
+    def _jenkins_container(self) -> Container:
+        """The Jenkins workload container."""
+        return self.unit.get_container("jenkins")
```
amandahla, line 57: What do you think about moving the container name to the state? The same was suggested [here](https://github.com/canonical/github-actions-exporter-operator/pull/3#discussion_r1197299662).

src/jenkins.py
```diff
@@ -3,38 +3,69 @@
...
-from ops.model import Container
 
-from types_ import Credentials, JenkinsEnvironmentMap
+import state
 
-JENKINS_WEB_URL = "http://localhost:8080"
-JENKINS_HOME_PATH = Path("/var/jenkins")
+logger = logging.getLogger(__name__)
+
+WEB_URL = "http://localhost:8080"
```
amandahla, line 23: Do you think 8080 should be a constant?

src/state.py
```diff
@@ -0,0 +1,107 @@
...
+            raise ValidationError(f"Fields {empty_fields} cannot be empty.")
+        try:
+            int(self.executors)
+        except ValueError as exc:
+            raise ValidationError(
+                f"Number of executors {self.executors} cannot be converted to type int."
+            ) from exc
+
+
+class State:
```
benhoyt, line 52: Would making this class a `@dataclass(frozen=True)` simplify? It would avoid the `__init__` and properties, and make it easier to document each attribute, in just a few lines of code. You'd still need `from_charm`, of course.

src/jenkins.py
```diff
@@ -47,17 +78,29 @@ def wait_jenkins_ready(timeout: int = 140, check_interval: int = 10) -> None:
...
         sleep(check_interval)
     else:
-        if _is_jenkins_ready():
+        if _is_ready():
             return
         raise TimeoutError("Timed out waiting for Jenkins to become ready.")
 
 
-def get_admin_credentials(connectable_container: Container) -> Credentials:
+class Credentials(typing.NamedTuple):
```
benhoyt, line 91: I think a `@dataclass` might be better for this, as it's not really a tuple.

yanksyoon: May I know why a `@dataclass` would be more suitable here? My understanding is that `tuple`s are immutable by default and that's what I would want when passing relevant group of data around. a `@dataclass` with `frozen=True` could do the same. Would you be able to help me understand what you mean by "as it's not really a tuple" please?
Thank you for the review!

benhoyt: Yeah, sorry, should have stated my motivation. I used to use namedtuple (the old-school one) all the time, so it's not *bad*, but the reason I consider data classes better for this kind of structure is that tuples are [sequence types](https://docs.python.org/3/library/stdtypes.html#typesseq), so are indexable by integer (but `credentials[0]` and `credentials[1]` doesn't really makes) and are "accidentally equal" to other tuples of the same length. See the ["Why not just use a namedtuple?"](https://peps.python.org/pep-0557/#why-not-just-use-namedtuple) section in PEP 557, the dataclasses proposal, for more info.

benhoyt: BTW, overall I think this PR is good stuff! I'm just thinking we should try to land on a few style things like "use dataclasses where appropriate" early on.

CC @jdkandersson: assuming you agree, maybe when you make updates to ISD014 you can show the State example with a dataclass too?

yanksyoon: This feedback was extremely helpful, I'll share it with the team as well. Thank you!

jdkandersson: Yeh I think dataclasses are also good, the spec should note that I agree

src/charm.py
```diff
@@ -6,36 +6,57 @@
...
 
-if TYPE_CHECKING:
+if typing.TYPE_CHECKING:
     from ops.pebble import LayerDict  # pragma: no cover
 
 
 logger = logging.getLogger(__name__)
 
 
+class AgentRelationData(typing.TypedDict):
```
jdkandersson, line 26: Should this be in an `Agent` module?

src/charm.py
```diff
@@ -6,36 +6,57 @@
...
 
-    def __init__(self, *args: Any):
+    def __init__(self, *args: typing.Any):
         """Initialize the charm and register event handlers.
 
         Args:
             args: Arguments to initialize the char base.
         """
         super().__init__(*args)
+        self.state = state.State.from_charm(self.model.config)
```
jdkandersson, line 48: Perhaps the import could be `from state import State` so that this doesn't have `state.State`?

src/charm.py
```diff
@@ -6,36 +6,57 @@
...
+
         self.framework.observe(self.on.jenkins_pebble_ready, self._on_jenkins_pebble_ready)
+        self.framework.observe(self.on["agent"].relation_joined, self._on_agent_relation_joined)
```
jdkandersson, line 51: Perhaps the observer pattern here?

src/charm.py
```diff
@@ -73,37 +94,112 @@ def _on_jenkins_pebble_ready(self, event: PebbleReadyEvent) -> None:
...
         )
         container.replan()
         try:
-            wait_jenkins_ready()
-            unlock_jenkins(container)
+            jenkins.wait_ready()
+            self.unit.status = MaintenanceStatus("Configuring Jenkins.")
+            jenkins.bootstrap(
+                container,
+                self.state.jnlp_port,
```
jdkandersson, line 117: Perhaps these could be grouped in some form so that this doesn't need to pass the specific items and rather make an object available that has all the required data available? That reduces the amount of coupling between charm.py and jenkins.py

src/charm.py
```diff
@@ -73,37 +94,112 @@ def _on_jenkins_pebble_ready(self, event: PebbleReadyEvent) -> None:
...
+                self.state.num_executors,
+                self.state.plugins,
+            )
             # Second Jenkins server start restarts Jenkins to bypass Wizard setup.
             container.add_layer(
                 "jenkins",
-                self._get_pebble_layer(calculate_env(admin_configured=True)),
+                self._get_pebble_layer(jenkins.calculate_env(admin_configured=True)),
```
jdkandersson, line 124: Do you think `admin_configured` is state as well?

src/charm.py
```diff
@@ -73,37 +94,112 @@ def _on_jenkins_pebble_ready(self, event: PebbleReadyEvent) -> None:
...
             self.unit.status = BlockedStatus("Timed out waiting for Jenkins.")
             return
+        except jenkins.JenkinsBootstrapError as exc:
+            logger.error("Error installing plugins, %s", exc)
+            self.unit.status = BlockedStatus("Error installling plugins.")
+            return
+
+        self.unit.status = ActiveStatus()
+
+    def _on_agent_relation_joined(self, event: RelationJoinedEvent) -> None:
```
jdkandersson, line 140: Observer pattern in an Agent class?

src/jenkins.py
```diff
@@ -68,12 +111,25 @@ def get_admin_credentials(connectable_container: Container) -> Credentials:
...
     user = "admin"
     password_file_contents = str(
-        connectable_container.pull(JENKINS_PASSWORD_FILE_PATH, encoding="utf-8").read()
+        connectable_container.pull(PASSWORD_FILE_PATH, encoding="utf-8").read()
     )
     return Credentials(username=user, password=password_file_contents.strip())
 
 
-def calculate_env(admin_configured: bool) -> JenkinsEnvironmentMap:
+class EnvironmentMap(typing.TypedDict):
```
jdkandersson, line 119: perhaps just `Environment`? The type already indicates it is a dict?

src/jenkins.py
```diff
@@ -105,3 +159,143 @@ def unlock_jenkins(connectable_container: Container) -> None:
...
+    """
+    _unlock_wizard(connectable_container)
+    _install_config(connectable_container, jnlp_port, num_executors)
+    try:
+        _install_plugins(connectable_container, plugins)
+    except JenkinsPluginError as exc:
+        raise JenkinsBootstrapError("Failed to bootstrap Jenkins.") from exc
+
+
+def get_client(client_credentials: Credentials) -> jenkinsapi.jenkins.Jenkins:
```
jdkandersson, line 241: What is the reason for making this a public method please?

tests/integration/test_core.py
jdkandersson, line 1: Any particular reason for calling this `test_core`?

tests/unit/test_charm.py
```diff
@@ -6,39 +6,70 @@
...
+    monkeypatch: pytest.MonkeyPatch,
+    raise_exception: Callable,
+):
+    """
+    arrange: given a patched jenkins bootstrap method that raises an exception.
+    act: when the jenkins_pebble_ready event is fired.
+    assert: the unit status should be in BlockedStatus.
+    """
+    # speed up waiting by changing default argument values
+    monkeypatch.setattr(jenkins.wait_ready, "__defaults__", (1, 1))
```
jdkandersson, line 59: This still doesn't seem optimal, not in this PR, in the background perhaps think about how this could be done differently?

src/charm.py
jdkandersson: From reading this, it seems to be getting the container object rather than the name?

yanksyoon: Fixed, thank you!

src/jenkins.py

src/state.py
yanksyoon: That did simplify the whole thing a lot! Fixed, thank you!

src/jenkins.py

.github/workflows/integration_test.yaml
```diff
@@ -7,3 +7,5 @@ jobs:
   integration-tests:
     uses: canonical/operator-workflows/.github/workflows/integration_test.yaml@main
     secrets: inherit
+    with:
+      tmate-debugging: true
```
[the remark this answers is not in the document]
yanksyoon: Removed, thanks for catching that!

config.yaml
```diff
@@ -0,0 +1,19 @@
...
+    default: 1
+    description: |
+      Number of executors to configure for jenkins server.
+  jnlp-port:
+    type: int
+    default: 48484
+    description: |
+      TCP port on which to listen for incoming connections from agents.
+      Set to 0 for a random port, or to -1 to disable JNLP entirely.
+  plugins:
```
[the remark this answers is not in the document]
yanksyoon: Removed plugins from the config options. Thank you!

config.yaml
```diff
@@ -0,0 +1,19 @@
+# Copyright 2023 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+options:
+  num-executors:
```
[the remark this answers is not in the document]
yanksyoon: Removed executors from the config options, replaced with a hard coded config in the template with the value of 0 (to disallow running on the server charm).
Thank you!

config.yaml
```diff
@@ -0,0 +1,19 @@
...
+    type: int
+    default: 1
+    description: |
+      Number of executors to configure for jenkins server.
+  jnlp-port:
+    type: int
+    default: 48484
```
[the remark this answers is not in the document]
yanksyoon: Defaulted to the standard jnlp port, thank you!

src/charm.py
yanksyoon: Refactored to agent observer module! Thank you!

src/charm.py
yanksyoon: Sounds good, fixed thank you!

src/charm.py
yanksyoon: Refactored to agent observer, thank you!

src/charm.py
yanksyoon: A lot of the previous states/config values have been removed, with `jnlp_port` being the only value remaining. Perhaps when there are more options available, it can be refactored to `BootstrapConfig`.

src/charm.py
yanksyoon: Refactored, thank you!

src/jenkins.py
yanksyoon: Fixed, thank you!

src/agent.py
```diff
@@ -0,0 +1,100 @@
...
+        """The Jenkins workload container."""
+        return self.charm.unit.get_container(self.state.jenkins_service_name)
+
+    def _on_agent_relation_joined(self, event: RelationJoinedEvent) -> None:
+        """Handle agent relation joined event.
+
+        Args:
+            event: The event fired from an agent joining the relationship.
+        """
+        if not (binding := self.model.get_binding("juju-info")):
```
jdkandersson, line 57: Under what conditions would the binding not be available?

src/agent.py
```diff
@@ -0,0 +1,100 @@
...
+            logger.error("Invalid agent relation data. %s", exc)
+            self.charm.unit.status = BlockedStatus("Invalid agent relation data.")
+            return
+
+        self.charm.unit.status = MaintenanceStatus("Adding agent node.")
+        credentials = jenkins.get_admin_credentials(self._jenkins_container)
+        jenkins_client = jenkins.get_client(client_credentials=credentials)
+        try:
+            jenkins.add_agent_node(
+                jenkins_client=jenkins_client,
```
jdkandersson, line 86: Can Jenkins figure out how to get its own client? I.e., just get `add_agent_node` to call `_get_client`?

src/jenkins.py
```diff
@@ -53,6 +52,46 @@ class JenkinsBootstrapError(JenkinsError):
...
+
+    def validate(self) -> None:
+        """Validate the agent metadata.
+
+        Raises:
+            ValidationError: if the field contains invalid data.
+        """
+        empty_fields = [
+            field
+            # Pylint doesn't understand that _fields is implemented in NamedTuple.
```
jdkandersson, line 81: Is this comment still accurate?

src/agent.py
yanksyoon: We would need to pass in the credentials, most of the jenkins functions would either need the credentials or the client. Passing in client seemed more explicit and easier to test.

jdkandersson: Couldn't those also just be requested by the client creation function?

Carrying around a Jenkins client and passing that around increases coupling between modules so it would be good to minimise if possible.

For testing, the jenkins client can be an optional argument

src/agent.py
yanksyoon: This was to remove the None type, updated to assertion, thank you!

src/agent.py
yanksyoon: Updated the client to be an optional argument! Thank you :)

tests/integration/test_core.py
yanksyoon: I'm matching other repositories (wordpress), to call it `test_core` since it tests the `core` functionalities. Other suggestions welcome!

jdkandersson: I think this contains agent tests? perhaps test_agent? That would make it much more obvious rather than core which could be open to interpretation

src/charm.py
yanksyoon: Refactored to `is_bootstrapped` function in the `jenkins` module :), thanks for the suggestion.
As of having the container for dynamic state, I think we could pass in the `self.unit` for the `from_charm` function to be able to call `get_container(self.jenkins_service_name)` within. 
The reason for it is to avoid deadlocking of `charm.py`'s `_jenkins_container` property
```
 @property
    def _jenkins_container(self) -> Container:
        """The Jenkins workload container."""
        return self.unit.get_container(self.state.jenkins_service_name)
```
and `State` initialization
```
self.state = State(..., self._jenkins_container)
```
(note that the `_jenkins_container` depends on the `state.jenkins_service_name` and the state would depend on `_jenkins_container`).

tests/integration/test_core.py
yanksyoon: Renamed, thank you!

jenkins_rock/rockcraft.yaml
```diff
@@ -36,12 +36,13 @@ parts:
...
-      - JENKINS_VERSION: 2.403
+      - JENKINS_VERSION: 2.387
+      - JENKINS_PLUGIN_MANAGER_VERSION: 2.12.11
     override-build: |
       mkdir -p ${CRAFT_PART_INSTALL}/{srv/jenkins/,etc/default/jenkins/}
-      chown 2000:2000 ${CRAFT_PART_INSTALL}/srv/jenkins
       # Use jenkins war rather than apt install for easier Jenkins version control.
       cd ${CRAFT_PART_INSTALL}/srv/jenkins
+      curl -sLO https://github.com/jenkinsci/plugin-installation-manager-tool/releases/download/${JENKINS_PLUGIN_MANAGER_VERSION}/jenkins-plugin-manager-${JENKINS_PLUGIN_MANAGER_VERSION}.jar
       curl -sLO https://mirrors.jenkins.io/war/${JENKINS_VERSION}/jenkins.war
```
gregory-schiano, line 46: Shouldn't we verify the war signature ?
See: https://www.jenkins.io/download/verify/

src/jenkins.py
```diff
@@ -105,3 +200,169 @@ def unlock_jenkins(connectable_container: Container) -> None:
...
+    """
+    _unlock_wizard(connectable_container)
+    _install_config(connectable_container, jnlp_port)
+    try:
+        _install_plugins(connectable_container)
+    except JenkinsPluginError as exc:
+        raise JenkinsBootstrapError("Failed to bootstrap Jenkins.") from exc
+
+
+def is_boostrapped(connectable_container: ops.Container) -> bool:
```
gregory-schiano, line 273: I'm wondering if we should cache this as a property, since once it's bootstrapped there are few chances that this would change

config.yaml
```diff
@@ -0,0 +1,8 @@
+# Copyright 2023 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+options:
+  jnlp-port:
+    type: int
+    default: 50000
+    description: TCP port on which to listen for incoming connections from agents.
```
gregory-schiano, line 8: What's the reasoning behind ? We should add any config param consciously.

tests/integration/test_agent.py
```diff
@@ -0,0 +1,56 @@
...
+
+    assert "Unlock Jenkins" not in str(response.content)
+    assert "Welcome to Jenkins!" in str(response.content)
+
+
+async def test_jenkins_agent_relation(
+    model: Model,
+    application: Application,
+    jenkins_k8s_agent: Application,
+    jenkins_client: jenkinsapi.jenkins.Jenkins,
```
gregory-schiano, line 36: I think we should have an integration test before that ensure the action is running properly to help in case of issues in the action that would fail this test
A simple call to jenkins api that would use this fixture before this test uses it

gregory-schiano (commented):
Stopped before unittests, will go on later

src/agent.py
```diff
@@ -0,0 +1,99 @@
...
+    Attrs:
+        url: The Jenkins server url.
+        secret: The secret for agent node.
+    """
+
+    url: str
+    secret: str
+
+
+class Observer(Object):
```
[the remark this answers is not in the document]
yanksyoon: I've received a feedback that importing as `agent.Observer` helps with clarity, since we can directly know where the import is coming from. Also, since we're importing from the module, it would be better not to repeat the name `Agent`. Would this be okay?

src/jenkins.py
yanksyoon: There's a plan to remove this function in favour of doing a `pebble.restart` :D

config.yaml
yanksyoon: I thought I had to maintain backwards compatibility, it was discussed that it doesn't need to be backwards compatible :), this will be hardcoded in the j2 template!

src/charm.py
yanksyoon: removed in favor of `container.restart`

tests/integration/test_agent.py
yanksyoon: The `jenkinsapi.jenkinis.Jenkins` on init, has a method that connects to the instance (hence the init would fail if there were any problems connecting to the instance). 
[Init](https://github.com/pycontribs/jenkinsapi/blob/ec2a23ff4d2ff539de34810da0a4189c6be1eb41/jenkinsapi/jenkins.py#L91) -> [poll] (https://github.com/pycontribs/jenkinsapi/blob/ec2a23ff4d2ff539de34810da0a4189c6be1eb41/jenkinsapi/jenkinsbase.py#LL40C18-L40C22) -> [python api request](https://github.com/pycontribs/jenkinsapi/blob/ec2a23ff4d2ff539de34810da0a4189c6be1eb41/jenkinsapi/jenkinsbase.py#L83).
Perhaps I'll document this behavior in the test so that it's less confusing! :D 
Thanks for the feedback!

jenkins_rock/rockcraft.yaml
yanksyoon: Added, thank you!

src/agent.py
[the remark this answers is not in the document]
gregory-schiano: That's ok as we don't have any explicit convention for this. Indeed using `agent.Observer` makes it clear, I just find it difficult when you're in a module to understand the context of an object only from the module name, at least for class names.

src/agent.py
```diff
@@ -0,0 +1,99 @@
...
+            )
+        except jenkins.JenkinsError as exc:
+            self.charm.unit.status = BlockedStatus(f"Jenkins API exception. {exc=!r}")
+            return
+
+        # This is to avoid the None type.
+        assert (binding := self.model.get_binding("juju-info"))  # nosec
+        host = binding.network.bind_address
+        event.relation.data[self.model.unit].update(
+            AgentRelationData(url=f"http://{str(host)}:8080", secret=secret)
```
amandahla, line 97: Shouldn't the constant WEB_PORT be used here instead of 8080?

src/jenkins.py
```diff
@@ -105,3 +192,139 @@ def unlock_jenkins(connectable_container: Container) -> None:
...
+            "-jar",
+            "jenkins-plugin-manager-2.12.11.jar",
+            "-w",
+            "jenkins.war",
+            "-d",
+            str(PLUGINS_PATH),
+            "-p",
+            plugins,
+        ],
+        working_dir="/srv/jenkins/",
```
amandahla, line 231: Just checking :)
HOME_PATH?

gregory-schiano (approved):
LGTM

src/jenkins.py
yanksyoon: The home path is `/var/jenkins` and `/srv/jenkins` is where the .war files are at. Perhaps I should put this as a constant too! Thanks for the comment!

src/agent.py
jdkandersson: I realise this got merged, doesn't the `f-string` apply `str` by default? So `AgentRelationData(url=f"http://{host}:8080", secret=secret)` does that work?

yanksyoon: Will update this on the next push!

PR #7, opened 2023-06-08
feat: auto update Jenkins

yanksyoon:
This PR addresses the management of Jenkins versions, namely:
1. Disable UI updates
2. Auto update Jar executable

[before UI disabling]
image
[after UI disabling]
image
[Configuration persist after restart (both /safeRestart and pebble restart)]
image

Out of scope:
- Ubuntu automatic updates
- Update time configuration option


src/charm.py
```diff
@@ -73,7 +75,7 @@ def _get_pebble_layer(self, jenkins_env: jenkins.Environment) -> Layer:
                 "online": {
                     "override": "replace",
                     "level": "ready",
-                    "http": {"url": jenkins.WEB_URL},
+                    "http": {"url": jenkins.LOGIN_URL},
```
yanksyoon, line 78: `WEB_URL` gives a redirect, `LOGIN_URL` will give 200 :)

src/jenkins.py
```diff
@@ -164,9 +183,11 @@ class Environment(typing.TypedDict):
 
     Attrs:
         JENKINS_HOME: The Jenkins home directory.
+        CASC_JENKINS_CONFIG: The Jenkins configuration-as-code plugin config path.
```
yanksyoon, line 186: This environment variable has to be named this way for JCasC to pick up.

yanksyoon (commented):
Some comments to help review :)

src/jenkins.py
```diff
@@ -212,14 +235,18 @@ def _unlock_wizard(connectable_container: ops.Container) -> None:
...
 
     Args:
         connectable_container: The connectable Jenkins workload container.
     """
     with open("templates/jenkins-config.xml", encoding="utf-8") as jenkins_config_file:
         connectable_container.push(CONFIG_FILE_PATH, jenkins_config_file, user=USER, group=GROUP)
+    with open("templates/jenkins.yaml", encoding="utf-8") as jenkins_casc_config_file:
+        connectable_container.push(
+            JCASC_CONFIG_FILE_PATH, jenkins_casc_config_file, user=USER, group=GROUP
+        )
```
weiiwang01, line 249: There's a method called `push_path` in `Container` class that can directly push a local file to the workload container.

yanksyoon: There currently seems to be a bug when using `push_path`, https://github.com/canonical/operator/issues/948 which causes the unit tests to fail. When the issue gets merged, I can change it! Thanks for the review!

weiiwang01 (approved):

src/jenkins.py
```diff
@@ -347,4 +375,129 @@ def add_agent_node(
...
+    """
+    try:
+        versions = _fetch_versions_from_rss()
+    except JenkinsNetworkError as exc:
+        raise JenkinsNetworkError("Failed to fetch LTS versions from RSS feed.") from exc
+
+    maj_min_version = _get_major_minor_version(current_version)
+    for version in versions:
+        if version.startswith(maj_min_version):
+            return version
```
gregory-schiano, line 454: This will return the first one matching, versions isn't sorted, are we sure the RSS feed will always return a sorted list ?
Wouldn't it be safer to get all the one that match and then return the highest patch version ?
I'm wondering if using a semver library could help as semver could be more that X.Y.Z

src/jenkins.py
```diff
@@ -347,4 +375,129 @@ def add_agent_node(
...
+    Args:
+        credentials: The credentials of a Jenkins user with access to the Jenkins API.
+        client: The API client used to communicate with the Jenkins server.
+
+    Raises:
+        JenkinsError: if there was an API error calling safe restart.
+    """
+    client = client if client is not None else _get_client(credentials)
+    try:
+        client.safe_restart(wait_for_reboot=True)
```
gregory-schiano, line 500: Does this method has a timeout ?
I don't think that's the case but any risk of parallel execution ? (shouldn't be the case since juju should execute 1 update-status event at a time)

tests/integration/helpers.py
gregory-schiano, line 1: We should avoid naming files `helpers` could you name in `substrings.py` for example ?

tests/unit/test_charm.py
```diff
@@ -131,3 +132,84 @@ def test__on_get_admin_password_action(
...
+    arrange: given unpatched jenkins, monkeypatched download_stable_war that raises an exception.
+    act: when update_status action is triggered.
+    assert: no action is taken and the charm remains active but the error is logged.
+    """
+    monkeypatch.setattr(jenkins, "get_version", lambda: versions.current)
+    monkeypatch.setattr(jenkins, "get_latest_patch_version", lambda *_, **__: versions.patched)
+    monkeypatch.setattr(
+        jenkins,
+        "download_stable_war",
+        lambda *_args, **_kwargs: raise_exception(exception=jenkins.JenkinsNetworkError),
```
gregory-schiano, line 176: `download_stable_war` could also raise a `ValidationError` which isn't covered, and probably work as expected

src/charm.py
```diff
@@ -130,6 +133,34 @@ def _on_get_admin_password(self, event: ActionEvent) -> None:
...
         event.set_results({"password": credentials.password})
 
+    def _on_update_status(self, _: UpdateStatusEvent) -> None:
+        """Handle update status event.
+
+        On Update status:
+        1. Update Jenkins patch LTS version if available.
+        2. Update apt packages if available.
+        """
+        self.unit.status = ActiveStatus("Checking for updates.")
```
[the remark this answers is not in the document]
yanksyoon: According to the [state values](https://discourse.charmhub.io/t/status-values/1168) document, `MaintenanceStatus` is "Getting ready, Not yet providing it's services", so I think the best choices here are either `WaitingStatus` or `ActiveStatus` but `WaitingStatus` is also not quite well suited "Healthy but waiting for another application to be Active, or a package installation".
Thanks for invoking the thought though!

src/charm.py
```diff
@@ -130,6 +133,34 @@ def _on_get_admin_password(self, event: ActionEvent) -> None:
...
+        try:
+            jenkins.download_stable_war(self._jenkins_container, latest_patch_version)
+            credentials = jenkins.get_admin_credentials(self._jenkins_container)
+            jenkins.safe_restart(credentials)
+        except (jenkins.JenkinsNetworkError, jenkins.JenkinsError) as exc:
+            error = "Failed to update Jenkins. See log for more details."
+            logger.error("Failed to update Jenkins. %s", exc)
+
+        self.unit.set_workload_version(latest_patch_version)
+        self.unit.status = ActiveStatus(error)
```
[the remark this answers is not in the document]
yanksyoon: My thought process was that even if there was an error updating Jenkins, as long as the service is still active, it doesn't quite categorize into other status.
For example, in the unlikely event of a corrupt RSS feed / executable download, it wouldn't affect the running Jenkins but a user's attention is required, hence the status message. This would be a nice case where a `warning` level of a status would fit in nicely but there currently isn't one.

src/jenkins.py
yanksyoon: From my understanding, the RSS feed was always sorted from latest to oldest, but yes, it would be beneficial to be safer, thanks for the suggestion!

src/jenkins.py
yanksyoon: Unfortunately it doesn't provide a timeout parameter, it's a polling method wrapped without a timeout. However, the polling is:
- every 5 seconds
- max 30 repeat
I think this is a sane configuration value that can be used :)
On the risk of parallel execution, there cannot be a way unless a user calls `/safeRestart` endpoint manually :(.

tests/unit/test_charm.py
yanksyoon: Added test cases for those! Thanks for catching that!

src/jenkins.py
```diff
@@ -347,4 +375,129 @@ def add_agent_node(
...
+        The latest patched version available.
+
+    Raises:
+        JenkinsNetworkError: if there was an error fetching the LTS RSS feed.
+        ValidationError: if the RSS feed contains no matching LTS version.
+    """
+    try:
+        versions = _fetch_versions_from_rss()
+    except JenkinsNetworkError as exc:
+        raise JenkinsNetworkError("Failed to fetch LTS versions from RSS feed.") from exc
```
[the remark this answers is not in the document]
yanksyoon: Fixed and added test cases, thanks for catching that!

src/charm.py
```diff
@@ -130,6 +133,34 @@ def _on_get_admin_password(self, event: ActionEvent) -> None:
...
+        self.unit.status = MaintenanceStatus("Updating Jenkins.")
+        error = ""
+        try:
+            jenkins.download_stable_war(self._jenkins_container, latest_patch_version)
+            credentials = jenkins.get_admin_credentials(self._jenkins_container)
+            jenkins.safe_restart(credentials)
+        except (jenkins.JenkinsNetworkError, jenkins.JenkinsError) as exc:
+            error = "Failed to update Jenkins. See log for more details."
+            logger.error("Failed to update Jenkins. %s", exc)
+
```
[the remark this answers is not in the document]
yanksyoon: Refactored this function to have more fine-grained set_status actions depending on the execution flow. Thanks for the review!

tests/integration/helpers.py
yanksyoon: This file contains status names that are used (to avoid `#type: ignore` everywhere) and a Exception raising patch class. `utils` is also not a great name, but it doesn't quite have things related to substrings. Perhaps should these be refactored to be in conftests?
One drawback I can think of when moving these to conftests is that the arguments for most of the tests are already ~5, the max without having to disable the linter for `too-many-arguments`.

src/jenkins.py
yanksyoon: I've tried refactoring with all versions using `semver` library. However, there were some caveats that I led me think that at this moment, it could be better to do string parsing.
1. Jenkins versions are not semantic - i.e. 2.387, we can parse this to `semver.Version` type. However, when converting it back to `str` (needed in most cases), it becomes `2.387.0`, without an option to omit the last patch version 0. This meant that whenever I need to convert the `semver.Version` back to `str`, the custom `version_to_string` method would need to follow everywhere. I'm not too sure if that would be beneficial.
2. The stable RSS feed would return  versions of format in x.y.z consistently. I think that we can rely on this not changing for a while.

tests/integration/conftest.py
```diff
@@ -163,3 +171,29 @@ async def jenkins_machine_agent_fixture(machine_model: Model) -> Application:
...
+    rss_feed_response = requests.get(jenkins.RSS_FEED_URL, timeout=10)
+    assert rss_feed_response.status_code == 200, "Failed to fetch RSS feed."
+    rss_xml = str(rss_feed_response.content, encoding="utf-8")
+    # extract all version strings from feed
+    pattern = r"\d+\.\d+\.\d+"
+    matches = re.findall(pattern, rss_xml)
+    # find first matching version starting with same <major>.<minor> version, the rss feed is
+    # sorted by latest first.
+    current_major_minor = ".".join(jenkins_version.split(".")[:2])
+    matched_latest_version = next((v for v in matches if v.startswith(current_major_minor)), None)
```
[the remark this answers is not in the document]
yanksyoon: Added sorting!

tests/integration/helpers.py
gregory-schiano: I only see `assert_substrings_not_in_string` in this module, am I missing something ?

yanksyoon: My bad, i was looking at the wrong file!

gregory-schiano (approved):
LGTM

jenkins_rock/rockcraft.yaml
```diff
@@ -46,7 +46,7 @@ parts:
       # Use jenkins war rather than apt install for easier Jenkins version control.
       cd ${CRAFT_PART_INSTALL}/srv/jenkins
       curl -sLO https://github.com/jenkinsci/plugin-installation-manager-tool/releases/download/${JENKINS_PLUGIN_MANAGER_VERSION}/jenkins-plugin-manager-${JENKINS_PLUGIN_MANAGER_VERSION}.jar
-      curl -sLO https://mirrors.jenkins.io/war/${JENKINS_VERSION}/jenkins.war
+      curl -sLO https://updates.jenkins.io/download/war/${JENKINS_VERSION}/jenkins.war
```
amandahla, line 49: Something that I remember to do today in Synapse rock: add a timeout to curl :) "-m" flag, I guess.

arturo-seijas (approved):
LGTM

amandahla (approved):
LGTM

PR #8, opened 2023-06-16
fix: workaround jenkinsapi safe_restart bug

yanksyoon:
There is a bug where in the dependent library `jenkinsapi` that raises an uncaught exception when `/safeRestart` is called on a Jenkins waiting for a job completion.

This PR addresses a workaround for it.

Relevant issue: https://github.com/pycontribs/jenkinsapi/issues/844
Relevant PR: https://github.com/pycontribs/jenkinsapi/pull/845

image
image


src/jenkins.py
```diff
@@ -114,6 +115,30 @@ def validate(self) -> None:
             ) from exc
 
 
+def _wait_for(func: typing.Callable, timeout: int = 300, check_interval: int = 10) -> None:
```
weiiwang01, line 118: The typing for `func` can be `typing.Callable[[], typing.Any]`

weiiwang01 (approved):

amandahla (approved):
LGTM

PR #9, opened 2023-06-20
feat: update-time-range

yanksyoon:
This PR enhances the charm by introducing a configuration `update-time-range` which allows charm users to configure the time range in which the charm can update the Jenkins patch versions.


config.yaml
```diff
@@ -0,0 +1,11 @@
+# Copyright 2023 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+options:
+  update-time-range:
+    type: string
+    description: >
+      A preferred update time in 24 hour format in UTC for the Jenkins LTS patch update to take
```
arturo-seijas, line 8: Keep in mind that we should aim to keep the grammar in the docs as simple as possible. Also, try not to use Latin terms over English ones. Some general guidelines can be found here https://developers.google.com/style/abbreviations. I wasn't able to find a topic covering this case in our documentation.

How about something in the lines of  `Preferred time for LTS patch updates in 24 hour UTC format. If empty, the update will take place whenever an update is available. For example, 03-05 will allow updates to take place from 3AM UTC to 5AM UTC.`

tests/unit/test_state.py
```diff
@@ -0,0 +1,108 @@
...
+
+import pytest
+
+import state
+
+
+@pytest.mark.parametrize(
+    "time_range",
+    [
+        pytest.param("-1-3", id="negative time range"),
```
arturo-seijas, line 16: Can you test for an invalid format too?

yanksyoon: These are all invalid test cases :)

arturo-seijas: I wasn't very clear :D I meant a string not matching the H(H)-H(H) pattern

src/state.py
```diff
@@ -4,25 +4,122 @@
 """Jenkins States."""
 import dataclasses
 import logging
+import typing
+from datetime import datetime
 
 logger = logging.getLogger(__name__)
 
 
+class CharmConfigInvalidError(Exception):
```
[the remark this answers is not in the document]
yanksyoon: I'm a bit conflicted about this, I've had the previous jenkins module separated out with `exceptions.py` and `types.py` and the feedback i've received was to have it within the module to allow for more cohesion, so that implementation was rejected. What would you think about a magic number of 3, where if we go over 3 of typing or exception within a module that the module becomes crowded, we can refactor it out to a separate module?
I think it's a good thing we can bring up to the team as well! 
Thank you for your feedback! :D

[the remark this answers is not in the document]
arturo-seijas: Maybe that was my feedback? It sounds like me. I know having files for several types/exceptions is common practice in Python, but I would rather strive for cohesion especially if these are only referred by a single class. e.g. a exception is only visible within one class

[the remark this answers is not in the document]
amandahla: I thought that would be a standard to follow since I saw that on Flask 🤔. I asked the same [here](https://github.com/canonical/github-runner-operator/pull/68/files#r1235389098).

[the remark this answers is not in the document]
yanksyoon: Yup, I think I additionally received the same comment from Ben during the external review :D

tests/unit/test_state.py
yanksyoon: Added the extra tests, thank you for the review!

config.yaml
yanksyoon: Updated the description, thank you for the guidelines!

src/state.py
```diff
@@ -4,25 +4,122 @@
...
+
+    Attrs:
+        start: Hour to allow updates from in UTC time, in 24 hour format.
+        end: Hour to allow updates until in UTC time, in 24 hour format.
+    """
+
+    start: int
+    end: int
+
+    def validate(self) -> None:
```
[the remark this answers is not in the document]
yanksyoon: After testing with the implementation with Pydantic, I'm not sure Pydantic is well suited for a class level validator (referred to as [root_validator](https://docs.pydantic.dev/latest/usage/validators/#root-validators) by pydantic).
I've found that the tooling that we use (pylint and mpy) will [complain](https://github.com/pydantic/pydantic/issues/568) heavily about the type hints, and to get over it, I would need to add either `# noqa` or `pylint: ignore=...` to get by.

Another thing is that the `@root_validator` will receive the values in dictionary format with `Optional[Any]` type. It's obvious(through static code checking) that the types will not be `None`. However, due to the typing provided in the `values` of the `@root_validator`,  I have to still check for `None` and the check for logic which adds complexity.

Lastly, the pydantic must be used with `no-name-in-module` [error ignored](https://github.com/canonical/flask-k8s-operator/blob/0421bfb3baad3b8ede7ce88583e089c779ee10b7/src/charm_state.py#L11).

The code below shows the implementation with Pydantic.
```
@root_validator
    @classmethod  # pylint doesn't understand that it is a class method.
    # https://github.com/pydantic/pydantic/issues/568
    def check_valid_time_range(cls, values: dict) -> dict:
        """Validate the time range.

        Args:
            values: The value keys of the model.

        Returns:
            The validated dictionary of values.

        Raises:
            ValueError: if the time range are out of bounds of 24H format.
        """
        start = values.get("start", None)
        end = values.get("end", None)
        if start is None or end is None:
            raise ValueError("Time range must not be empty.")
        if start < 0 or start > 23 or end < 0 or end > 23:
            raise ValueError("Time range out of 24 hour bounds.")
        if start == end:
            raise ValueError("Time range cannot be equal. Minimum 1 hour range is required.")
        return values
 ```

tl;dr;
Pydantic is very useful for field level checking, but perhaps might be an overkill for a class level validation, especially given that the field level checking is done through static code analysis and doesn't quite need special enhancements such as regex.

Happy to still change the implementation to Pydantic if we decide to use it as a standard :)

Thank you for the comment!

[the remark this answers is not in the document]
Comment (amandahla) [reply src/state.py:40]: @jdkandersson Please, could you give us help here? What do you think?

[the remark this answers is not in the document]
Comment (jdkandersson) [reply src/state.py:40]: yes please use pydantic as discussed on the stand up

File: src/state.py (hunk 3 above)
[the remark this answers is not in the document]
Comment (jdkandersson) [reply src/state.py:13]: Exceptions should inherit from a single parent for a given charm so that it is easy to handle all exceptions for the charm

Regarding where to put them, in packages it can be useful because the file that defines all exceptions is useful for the users of that package to know all the exceptions that can be raised. In charms it is probably ok to leave them in a module

File: src/state.py (hunk 4 above)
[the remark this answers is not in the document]
Comment (yanksyoon) [reply src/state.py]: Changed to use Pydantic, thank you for the reviews!

File: .woke.yaml [hunk 5]
```diff
@@ -1,3 +1,5 @@
 rules:
   # Ignore "slave" - the relation interface name for agents use this terminology.
   - name: slave
+  # Ignore whitelist - we are using it to ignore pydantic in pyptoject.toml
+  - name: whitelist
```
arturo-seijas, line 5: Missing blank line at EOF

yanksyoon: Woops, fixed! Thank you!

arturo-seijas (dismissed):
LGTM

tests/integration/test_jenkins.py
```diff
@@ -29,6 +32,94 @@ async def test_jenkins_update_ui_disabled(
...
+    time.sleep(15)
+    await model_app_unit.model.wait_for_idle(status="active")
+
+    model_status: client.FullStatus = await model_app_unit.model.get_status()
+    app_status = model_status.applications.get(model_app_unit.app.name)
+    assert app_status, "application status not found."
+
+    assert app_status.workload_version == jenkins_version
+
+    # Reset charm.py & model hook interval
```
jdkandersson, line 117: Should this be done in a fixture? This code won't get executed on a failure

tests/unit/test_state.py
```diff
@@ -0,0 +1,112 @@
...
```
jdkandersson, line 16: Should the edges be probed, i.e., -1, 0, 1, 22, 23, 24? Off by one is a common mistake that can be caught by these kind of edge tests

tests/unit/test_state.py
```diff
@@ -0,0 +1,112 @@
...
+    assert update_time_range, "Expected time range to not be None."
+
+    assert update_time_range.start == expected_range[0], "Unexpected start time."
+    assert update_time_range.end == expected_range[1], "Unexpected end time."
+
+
+@pytest.mark.parametrize(
+    "patch_hour,time_range,expected_result",
+    [
+        pytest.param(3, "3-5", True, id="start hour"),
```
jdkandersson, line 79: What about just before?

tests/integration/test_jenkins.py
```diff
@@ -29,6 +32,94 @@ async def test_jenkins_update_ui_disabled(
...
+        )
+        assert not ret, f"failed to scp modified charm, {stderr}"
+        (ret, _, stderr) = await ops_test.juju(
+            "ssh", unit_name, "/usr/bin/chmod", "755", f"{charm_path}/src/charm.py"
+        )
+        assert not ret, f"failed to chmod modified charm, {stderr}"
+        (ret, _, stderr) = await ops_test.juju(
+            "ssh", unit_name, "/usr/bin/chown", "root:root", f"{charm_path}/src/charm.py"
+        )
+        assert not ret, f"failed to chown modified charm, {stderr}"
```
gregory-schiano, line 104: I'm not so fan of this way of doing. There's a library "libfaketime" that can normally be easily injected using LD_PRELOAD envvar
I'm trying to figure out if there's a way to pass any environment variable to the charm context that would enable this solution

tests/unit/test_state.py
gregory-schiano: I don't get your comments 
`-1` is tested here
`0`, `1`, `22`, `23` aren't invalid
`24` is tested on the next param

tests/unit/test_state.py
gregory-schiano: And after a past midnight should be added too
`pytest.param(2, "23-1", False, id="out of overnight range"),`

tests/integration/test_jenkins.py
gregory-schiano: Ok I've 2 proposed approaches:
- Instead of trying to messup with the time, we can compute the time range based on the current hours (if it's 9AM and we want an interval that leads to an update then we do `f"{now().hour-1}-{now().hour+1}"` and if we want an interval that doesn't trigger an update `f"{now().hour+1}-{now().hour-1}"`
- I tested successfuly the following process:
```bash
juju run --unit traefik-k8s/0 -- apt update
juju run --unit traefik-k8s/0 -- apt install -y libfaketime
ubuntu@charm-dev:~$ juju run --unit traefik-k8s/0 -- 'LD_PRELOAD="/usr/lib/aarch64-linux-gnu/faketime/libfaketime.so.1" FAKETIME="@2000-01-01 11:12:13" date'
```
Which gives the following output:
`Sat Jan  1 11:12:13 UTC 2000`

Now if I run on a modified traefik charm that logs the current time:
```bash
juju run --unit traefik-k8s/0 -- 'LD_PRELOAD="/usr/lib/aarch64-linux-gnu/faketime/libfaketime.so.1" FAKETIME="@2000-01-01 11:12:13" JUJU_DISPATCH_PATH=hooks/upgrade-charm JUJU_MODEL_NAME=dqlite JUJU_UNIT_NAME=traefik-k8s/0 ./dispatch'
```

I'll get in the logs:
`unit-traefik-k8s-0: 17:47:32 INFO unit.traefik-k8s/0.juju-log 2000-01-01 11:12:14.219210`

tests/unit/test_state.py
jdkandersson: I didn't mean exactly here, across the tests (both invalid and valid cases)

config.yaml
```diff
@@ -0,0 +1,11 @@
+# Copyright 2023 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+options:
+  update-time-range:
+    type: string
+    description: >
+      Preferred UTC time range for LTS patch updates in 24 hour  format. If empty, the update will
+      take place whenever an update is available. For example, 03-05 will allow updates to take
```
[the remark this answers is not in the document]
yanksyoon: Added explanation on running job, thank you!

src/charm.py
```diff
@@ -145,15 +149,19 @@ def _on_get_admin_password(self, event: ActionEvent) -> None:
         credentials = jenkins.get_admin_credentials(self._jenkins_container)
         event.set_results({"password": credentials.password})
 
-    def _on_update_status(self, _: UpdateStatusEvent) -> None:
+    # The controller function has many branching logic and hence the C901 - too complex is ignored.
```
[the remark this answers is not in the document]
yanksyoon: Refactored the control flow for fetching current and latest version to a wrapper `get_updatable_version`, which returns None if latest version is already applied. Thank you!

src/state.py
```diff
@@ -4,25 +4,137 @@
...
 import logging
+import typing
+from datetime import datetime
+
+from pydantic import BaseModel, ValidationError, root_validator
 
 logger = logging.getLogger(__name__)
 
 
+class CharmConfigInvalidError(Exception):
```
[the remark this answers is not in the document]
yanksyoon: Added `CharmStateBaseError`, thank you!

src/state.py
```diff
@@ -4,25 +4,137 @@
...
+        Returns:
+            A dictionary validated values.
+
+        Raises:
+            ValueError: if the time range are out of bounds of 24H format.
+        """
+        # it is okay to cast it since the field level validation has ran before root validation.
+        start = typing.cast(int, values["start"])
+        end = typing.cast(int, values["end"])
+        if start < 0 or start > 23 or end < 0 or end > 23:
```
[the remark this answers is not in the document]
yanksyoon: Added as individual Fields using pydantic, thank you!

src/state.py
```diff
@@ -4,25 +4,137 @@
...
+            None: if the input time range value is None or empty.
+        """
+        if not time_range:
+            return None
+        try:
+            (start_hour, end_hour) = (int(hour) for hour in time_range.split("-"))
+            update_range = cls(start=start_hour, end=end_hour)
+        except (ValueError, ValidationError) as exc:
+            raise ValueError(
+                f"Invalid time range {time_range}, time range must be in 24H format"
```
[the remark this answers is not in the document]
yanksyoon: Refactored the functions to have separate flow for ValueError and ValidationError. The `TimeRange` class is refactored to a separate module. The state will only handle the error coming from the module.

src/state.py
```diff
@@ -4,25 +4,137 @@
...
+            ) from exc
+        return update_range
+
+    def check_now(self) -> bool:
```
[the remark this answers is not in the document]
yanksyoon: Refactored into a separate module, thank you!

src/state.py
```diff
@@ -4,25 +4,137 @@
...
+            UpdateTimeRange: if a valid time range was given.
+            None: if the input time range value is None or empty.
+        """
+        if not time_range:
+            return None
+        try:
+            (start_hour, end_hour) = (int(hour) for hour in time_range.split("-"))
+            update_range = cls(start=start_hour, end=end_hour)
+        except (ValueError, ValidationError) as exc:
+            raise ValueError(
```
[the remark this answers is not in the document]
yanksyoon: Introduced another state specific error `CharmConfigInvalidError`, thank you!

tests/integration/test_jenkins.py
yanksyoon: I think the environment variable solution would be a good thing to try here so that when other charms face similar issues, the current repository can be referred to for manipulating time. Thank you for your research & suggestion!

src/state.py
```diff
@@ -4,25 +4,137 @@
...
+            True if within bounds, False otherwise.
+        """
+        current_hour = datetime.utcnow().time().hour
+        if self.start <= current_hour < self.end:
+            return True
+        if self.end < self.start:
+            if self.end <= current_hour < self.start:
+                return False
+            return True
+        return False
```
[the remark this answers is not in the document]
yanksyoon: Fixed, thank you!

tests/integration/test_jenkins.py
yanksyoon: Implemented to manually trigger update-status-hook using libfaketime environment variables.

tests/integration/test_jenkins.py
yanksyoon: Refactored to fixture, thank you for the suggestion!

tests/integration/test_jenkins.py
```diff
@@ -29,6 +32,94 @@ async def test_jenkins_update_ui_disabled(
...
+    act: when update-time-range between 3AM to 5AM is applied.
+    assert: the update does not take place.
+    """
+    await model_app_unit.model.wait_for_idle()
+    await model_app_unit.app.set_config({"update-time-range": "03-05"})
+    unit_name: str = model_app_unit.unit.name
+    charm_path = f"/var/lib/juju/agents/unit-{unit_name.replace('/', '-')}/charm"
+
+    # Overwrite requirements.txt
+    with tempfile.NamedTemporaryFile("wt", encoding="utf-8") as temp_requirements_txt, open(
```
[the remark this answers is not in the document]
yanksyoon: Refactored to manually trigger update-status-hook, thank you for the comment!

tests/unit/test_state.py
yanksyoon: Updated the test cases to include the before/edge/during/after cases :D Thanks for the review!

tests/unit/test_state.py
yanksyoon: Updated the test cases to include the before/edge/during/after cases :D Thanks for the review!

jdkandersson:
PR is looking pretty good

src/timerange.py
```diff
@@ -0,0 +1,87 @@
...
+    def check_now(self) -> bool:
+        """Check whether the current time is within the defined bounds.
+
+        Returns:
+            True if within bounds, False otherwise.
+        """
+        current_hour = datetime.utcnow().time().hour
+        # If the range crosses midnight
+        if self.start > self.end:
+            return self.start <= current_hour or current_hour < self.end
```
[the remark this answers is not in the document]
yanksyoon: Switched, thank you!

src/state.py
```diff
@@ -114,27 +43,32 @@ class State:
         update_time_range: Time range to allow Jenkins to update version.
     """
 
-    update_time_range: typing.Optional[UpdateTimeRange]
+    update_time_range: typing.Optional[Range]
     jenkins_service_name: str = "jenkins"
 
     @classmethod
-    def from_charm(cls, config: typing.Mapping[str, str]) -> "State":
+    def from_charm(cls, charm: "JenkinsK8sOperatorCharm") -> "State":
```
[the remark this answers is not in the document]
yanksyoon: Switched to CharmBase, thank you!

jdkandersson (approved):

arturo-seijas (approved):
LGTM

PR #10, opened 2023-06-28
extract docs for modules

jdkandersson:
After experimentation and feedback, settled on lazydocs. Going forward, the pre-commit hook that generates the src docs should be enabled so that the docs are automatically generated.


src/agent.md
```diff
@@ -0,0 +1,38 @@
...
+    * builtins.dict
+
+    ### Class variables
+
+    `secret: str`
+    :
+
+    `url: str`
+    :
+
```
yhaliaw, line 26: Is there any changes we can make to make the documentation for these two variable show up here?

lazydocs/agent.md
```diff
@@ -0,0 +1,66 @@
+<!-- markdownlint-disable -->
+
+<a href="../src/agent.py#L0"><img align="right" style="float:right;" src="https://img.shields.io/badge/-source-cccccc?style=flat-square"></a>
+
+# <kbd>module</kbd> `agent`
+The Jenkins agent relation observer.
```
jdkandersson, line 6: test

yanksyoon (approved):
LGTM!

tox.ini
```diff
@@ -29,6 +29,18 @@ commands =
...
+[testenv:src-docs]
+allowlist_externals=sh
+setenv =
+    PYTHONPATH=src
+description = Generate documentation for src
+deps =
+    lazydocs
+    -r{toxinidir}/requirements.txt
+commands =
+    ; can't run lazydocs directly due to needing to run it on src/* which produces an invocation error in tox
```
gregory-schiano, line 41: For my understanding, why should it run on "src/*" and not `src/` ?

jdkandersson: This is a bit messy because `lazydocs` expects a module whereas the charm isn't really a module (e.g., it doesn't have an `__init__.py`). To make `lazydocs` work, it is necessary to run from the root directory and add `src` to the `PYTHONPATH`. This results in pydantic seeing the `timerange.py` twice which results in an error around pydantic not allowing re-using root validators even though the code, as running in production, doesn't do this. One option is to add an argument to allow re-use, though it isn't required. This way, each module is treated independently rather than collectively, which is also more like what the charm does in production

jdkandersson: I spent a few hours trying to make this work, this is the best I could figure out

README.md
```diff
@@ -33,3 +33,12 @@ Thinking about using the Jenkins-k8s Operator for your next project?
...
 [see the charm's detailed documentation](https://charmhub.io/jenkins-k8s/docs).
+
+## Generating src docs for every commit
+
+Run the following command:
+
+```bash
+echo -e "tox -e src-docs\ngit add src-docs\n" > .git/hooks/pre-commit
+chmod +x .git/hooks/pre-commit
+```
```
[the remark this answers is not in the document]
jdkandersson: `pre-commit` is quite good for running tooling that it supports, like black, isort etc. It doesn't work nicely with tox since tox maintains its own dependencies which pre-commit also wants to do. We're pretty deep in tox so pre-commit is not a very good fit for us (maybe things have changed since I last tried)

gregory-schiano (approved):
LGTM

PR #11, opened 2023-07-11
feat: new agent_v0_interface support

yanksyoon:
This PR introduces the new agent_v0_interface under the `agent` relation.

It further addresses:
- code styling changes to import directly from the `ops` pacakge.
- removing `self._jenkins_container` replacing it as single use container get
- 1:1 callback methods for each event handler
- remove registered agent on relation broken.

### DIFF
--- a/metadata.yaml
+++ b/metadata.yaml
@@ -38,7 +38,7 @@ resources:
     type: oci-image
     description: OCI image for Jenkins
 requires:
-  slave:
+  agent-deprecated:
     interface: jenkins-slave
     optional: true
   agent:
--- a/metadata.yaml
+++ b/metadata.yaml
@@ -38,6 +38,9 @@ resources:
     type: oci-image
     description: OCI image for Jenkins
 requires:
-  agent:
+  slave:
     interface: jenkins-slave
+    optional: true
+  agent:
+    interface: jenkins_agent_v0
     optional: true
--- a/src-docs/agent.py.md
+++ b/src-docs/agent.py.md
@@ -5,6 +5,10 @@
 # <kbd>module</kbd> `agent.py`
 The Jenkins agent relation observer. 
 
+**Global Variables**
+---------------
+- **AGENT_RELATION**
+- **SLAVE_RELATION**
 
 
 ---
@@ -28,7 +32,7 @@ Relation data required for adding the Jenkins agent.
 ## <kbd>class</kbd> `Observer`
 The Jenkins agent relation observer. 
 
-<a href="../src/agent.py#L33"><img align="right" style="float:right;" src="https://img.shields.io/badge/-source-cccccc?style=flat-square"></a>
+<a href="../src/agent.py#L31"><img align="right" style="float:right;" src="https://img.shields.io/badge/-source-cccccc?style=flat-square"></a>
 
 ### <kbd>function</kbd> `__init__`
 
--- a/src-docs/agent.py.md
+++ b/src-docs/agent.py.md
@@ -8,7 +8,7 @@ The Jenkins agent relation observer.
 **Global Variables**
 ---------------
 - **AGENT_RELATION**
-- **SLAVE_RELATION**
+- **DEPRECATED_AGENT_RELATION**
 
 
 ---
--- a/src-docs/jenkins.py.md
+++ b/src-docs/jenkins.py.md
@@ -204,7 +204,7 @@ Add a Jenkins agent node.
 
 ---
 
-<a href="../src/jenkins.py#L499"><img align="right" style="float:right;" src="https://img.shields.io/badge/-source-cccccc?style=flat-square"></a>
+<a href="../src/jenkins.py#L522"><img align="right" style="float:right;" src="https://img.shields.io/badge/-source-cccccc?style=flat-square"></a>
 
 ## <kbd>function</kbd> `get_updatable_version`
 
@@ -228,7 +228,7 @@ Get version to update to if available.
 
 ---
 
-<a href="../src/jenkins.py#L525"><img align="right" style="float:right;" src="https://img.shields.io/badge/-source-cccccc?style=flat-square"></a>
+<a href="../src/jenkins.py#L548"><img align="right" style="float:right;" src="https://img.shields.io/badge/-source-cccccc?style=flat-square"></a>
 
 ## <kbd>function</kbd> `download_stable_war`
 
@@ -254,7 +254,7 @@ Download and replace the war executable.
 
 ---
 
-<a href="../src/jenkins.py#L585"><img align="right" style="float:right;" src="https://img.shields.io/badge/-source-cccccc?style=flat-square"></a>
+<a href="../src/jenkins.py#L608"><img align="right" style="float:right;" src="https://img.shields.io/badge/-source-cccccc?style=flat-square"></a>
 
 ## <kbd>function</kbd> `safe_restart`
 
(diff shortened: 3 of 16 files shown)


metadata.yaml
```diff
@@ -38,6 +38,9 @@ resources:
     type: oci-image
     description: OCI image for Jenkins
 requires:
-  agent:
+  slave:
     interface: jenkins-slave
```
mthaddon, line 42: Is it required to support this relation? We're trying to remove references to non-inclusive words in our charms, so it would be good if we could avoid this if possible (presumably as long as the jenkins-agent and jenkins-k8s charms talk the same relation that would be okay?)

metadata.yaml
```diff
@@ -38,6 +38,9 @@ resources:
...
     optional: true
+  agent:
+    interface: jenkins_agent_v0
```
mthaddon, line 45: This is the first time I've seen a version number appended to a relation. It's an interesting pattern, wonder if it's been discussed? Another approach could be for instance to include a version in the relation data.

src-docs/agent.py.md
```diff
@@ -5,6 +5,10 @@
 # <kbd>module</kbd> `agent.py`
 The Jenkins agent relation observer. 
 
+**Global Variables**
+---------------
+- **AGENT_RELATION**
+- **SLAVE_RELATION**
```
mthaddon, line 11: Same comment as above about non-inclusive words if possible.

mthaddon (commented):
Some questions about if it's possible to avoid non-inclusive words, and the interface name

metadata.yaml
yanksyoon: @jdkandersson shall I drop support for the old relation since there isn't a strict backwards compatible requirement?

Otherwise, I could suggest that I just rename this relation (keeping the interface) to something like `agent-deprecated` so that the old interface could be kept. Thank you for the review!

src-docs/agent.py.md
yanksyoon: Renamed to `agent-deprecated`!

metadata.yaml
jdkandersson: We should maintain support for the previous relation until we have done the production migrations to make them easier (one side of the relation can be upgraded and then the other rather than needing to do both at the same time)

src-docs/jenkins.py.md
```diff
@@ -204,7 +204,38 @@ Add a Jenkins agent node.
...
-<a href="../src/jenkins.py#L499"><img align="right" style="float:right;" src="https://img.shields.io/badge/-source-cccccc?style=flat-square"></a>
+<a href="../src/jenkins.py#L416"><img align="right" style="float:right;" src="https://img.shields.io/badge/-source-cccccc?style=flat-square"></a>
+
+## <kbd>function</kbd> `remove_agent_node`
+
+```python
+remove_agent_node(
+    agent_name: str,
+    credentials: Credentials,
+    client: Jenkins | None = None
```
jdkandersson, line 215: under what circumstances would this be `None`?

src/agent.py
```diff
@@ -96,4 +101,112 @@ def _on_agent_relation_joined(self, event: RelationJoinedEvent) -> None:
...
+
+        Args:
+            event: The event fired from an agent joining the relationship.
+        """
+        container = self.charm.unit.get_container(self.state.jenkins_service_name)
+        if not container.can_connect():
+            event.defer()
+            return
+        if not event.unit or not all(
+            event.relation.data[event.unit].get(required_data)
```
jdkandersson, line 117: Should the relation data be part of the state and validated using pydantic?

src/agent.py
```diff
@@ -96,4 +101,112 @@ def _on_agent_relation_joined(self, event: RelationJoinedEvent) -> None:
...
+            agent_meta.validate()
+        except jenkins.ValidationError as exc:
+            logger.error("Invalid agent relation data. %s", exc)
+            self.charm.unit.status = ops.BlockedStatus("Invalid agent relation data.")
+            return
+
+        self.charm.unit.status = ops.MaintenanceStatus("Adding agent node.")
+        credentials = jenkins.get_admin_credentials(container)
+        try:
+            jenkins.add_agent_node(
```
jdkandersson, line 138: Couldn't this function just take the container and get its own credentials? Would reduce the coupling

src/agent.py
```diff
@@ -96,4 +101,112 @@ def _on_agent_relation_joined(self, event: RelationJoinedEvent) -> None:
...
+
+        # The relation data is removed before this particular hook runs, making the name set by the
+        # agent not available. Hence, we can try to infer the name of the unit.
+        # See discussion: https://github.com/canonical/operator/issues/888
+        # assert type since event unit cannot be None.
+        agent_name = jenkins.get_agent_name(typing.cast(ops.Unit, event.unit).name)
+        credentials = jenkins.get_admin_credentials(container)
+        self.charm.unit.status = ops.MaintenanceStatus("Removing agent node.")
+        try:
+            jenkins.remove_agent_node(agent_name=agent_name, credentials=credentials)
```
jdkandersson, line 177: Similar to above, can this function figure out the `agent_name` and credentials itself? In general, try and limit the number of interactions from the charm classes to the modules. If there are multiple function calls - it is probably worth creating a higher level function on the module that does what the multiple function calls do

src-docs/jenkins.py.md
yanksyoon: the client is `None` if it should be created using the credentials passed in. It is not `None` during tests, where the client is injected.

src/agent.py
yanksyoon: Refactored to be in state, thank you for the review!

src/agent.py
yanksyoon: Refactored, thank you!

src/agent.py
yanksyoon: Refactored, thank you!

src/agent.py
```diff
@@ -41,59 +39,156 @@ def __init__(self, charm: CharmBase, state: State):
...
-        if not event.unit or not all(
-            event.relation.data[event.unit].get(required_data)
-            for required_data in ("executors", "labels", "slavehost")
-        ):
+        # The relation is joined, it cannot be None, hence the type casting.
+        deprecated_agent_relation_meta = typing.cast(
+            typing.Mapping[ops.Unit, AgentMeta], self.state.deprecated_agent_relation_meta
+        )
+        # The event unit cannot be None.
+        agent_meta = deprecated_agent_relation_meta[typing.cast(ops.Unit, event.unit)]
```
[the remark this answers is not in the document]
yanksyoon: This is taken from the `RelationData` class from`ops` `model.py`, where the mapping definition is `Mapping['UnitOrApplication', 'RelationDataContent'])`
```
Each unit and application involved in a relation has their own data bucket.
Eg: ``{entity: RelationDataContent}``
where entity can be either a :class:`Unit` or a :class:`Application`.
```
Would you recommend changing the key to perhaps something like `unit.name`?

src/agent.py
```diff
@@ -41,59 +39,156 @@ def __init__(self, charm: CharmBase, state: State):
...
+    def _on_deprecated_agent_relation_departed(self, event: ops.RelationDepartedEvent) -> None:
+        """Handle deprecated agent relation departed event.
+
+        Args:
+            event: The event fired when a unit in deprecated agent relation is departed.
+        """
+        # the event unit cannot be None.
+        container = self.charm.unit.get_container(self.state.jenkins_service_name)
+        if not container.can_connect():
+            return
```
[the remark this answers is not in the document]
yanksyoon: Good catch thank you!

src/agent.py
```diff
@@ -41,59 +39,156 @@ def __init__(self, charm: CharmBase, state: State):
...
+    def _on_agent_relation_departed(self, event: ops.RelationDepartedEvent) -> None:
+        """Handle agent relation departed event.
+
+        Args:
+            event: The event fired when a unit in agent relation is departed.
+        """
+        # the event unit cannot be None.
+        container = self.charm.unit.get_container(self.state.jenkins_service_name)
+        if not container.can_connect():
+            return
```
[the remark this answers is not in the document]
yanksyoon: Good catch thank you!

src/agent.py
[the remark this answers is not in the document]
weiiwang01: Well, if the relation data comes from the framework, I think there's nothing we can do. But I do think using the `unit.name` string could be a better solution.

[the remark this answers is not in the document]
yanksyoon: Oh I meant that the mapping of unit to data paradigm was taken from the ops library. Updated the key to be in `str` using the `unit.name`. Thank you for the suggestion! Makes it much more solid!

weiiwang01 (approved):

src/agent.py
```diff
@@ -113,33 +104,24 @@ def _on_agent_relation_joined(self, event: ops.RelationJoinedEvent) -> None:
...
-        if not event.unit or not all(
-            event.relation.data[event.unit].get(required_data)
-            for required_data in ("executors", "labels", "name")
-        ):
+        # The relation is joined, it cannot be None, hence the type casting.
+        agent_relation_meta = typing.cast(
+            typing.Mapping[str, AgentMeta], self.state.agent_relation_meta
+        )
+        # The event unit cannot be None.
+        agent_meta = agent_relation_meta[typing.cast(ops.Unit, event.unit).name]
```
jdkandersson, line 112: @benhoyt interesting case, if the unit cannot be None should the type hint be changed?

jdkandersson (approved):

src/agent.py
benhoyt: Yeah, there's almost certainly room for improvement there. The doc comment says about `RelationEvent.unit`:

```
     This will be ``None`` if the relation event was triggered as an
    :class:`Application <model.Application>`-level event.
```

I don't actually know what that means!

I also have a comment on `RelationEvent.app` that says, "# TODO(benhoyt): I *think* app should never be None, but confirm and update type".

It's quite possible the more specific event types, like `RelationJoinedEvent`, can have narrower attribute types. Feel free to open a `canonical/operator` issue to look into improving this.

yanksyoon: Can do!

PR #12, opened 2023-07-14
Test/agent relation integration

yanksyoon:
This PR implements integration test for recent changes in Jenkins-agent machine charm (agent_v0_interface support).

### DIFF
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -24,6 +24,7 @@ from pytest import FixtureRequest
 from pytest_operator.plugin import OpsTest
 
 import jenkins
+import state
 
 from .types_ import ModelAppUnit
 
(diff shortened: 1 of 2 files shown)


jdkandersson (approved):

tests/integration/conftest.py
```diff
@@ -185,16 +227,73 @@ async def machine_model_fixture(
...
+) -> typing.AsyncGenerator[Application, None]:
+    """The jenkins machine agent with 3 units."""
+    # 2023-06-02 use the edge version of jenkins agent until the changes have been promoted to
+    # stable.
+    app = await machine_model.deploy(
+        "jenkins-agent",
+        channel="latest/edge",
+        config={"labels": "machine"},
+        application_name=f"jenkins-agent-{app_suffix}",
+        num_units=num_units,
```
gregory-schiano, line 263: Its the same fixture as the previous one but with num_units specified, is there a way to parametrize this ?

yanksyoon: The multi-unit fixture is used for the new relation, multi-units for old relation will fail, hence the differentiation without parametrization. 
Shall I rename this fixture to perhaps something like `new_relation_agent`?

tests/integration/test_agent.py
```diff
@@ -123,16 +125,72 @@ async def test_jenkins_k8s_agent_relation(
...
+    arrange: given a cross controller cross model jenkins machine agent with an offer.
+    act: when the relation is setup through the offer.
+    assert: the relation succeeds and the agent is able to run jobs successfully.
+    """
+    machine_model: Model = jenkins_multi_machine_agents.model
+    await application.relate(
+        state.AGENT_RELATION,
+        f"localhost:admin/{machine_model.name}.{jenkins_multi_machine_agents.name}",
+    )
+    await model.wait_for_idle(apps=[application.name], status="active")
```
amandahla, line 162: Minor/no need to change: There is a wait_for_active parameter that can be used if you want all the units to be active.

amandahla (dismissed):
LGTM

tests/integration/test_agent.py
yanksyoon: changed! quite neat! thank you for the review!

tests/integration/conftest.py
yanksyoon: Changed! Thank you for the review!

jdkandersson (approved):

PR #18, opened 2023-07-20
feat: proxy config

yanksyoon:
This PR adds proxy configuration options required for Jenkins to work behind proxied environments.

Since adding the integration tests cost outweigh the benefits, attaching the screenshot of the feature working.

image
[Image of correctly installed jenkins through squid proxy, squid proxy access logs]
image
[Relevant squid proxy config]


config.yaml
```diff
@@ -9,3 +9,37 @@ options:
       take place whenever an update is available. For example, 03-05 will allow updates to take
       place from 3AM UTC to 5AM UTC. Awaits for running job completion for 5 minutes. 
     default: ""
+  proxy-hostname:
```
arturo-seijas, line 12: Shouldn't be fetching these from the juju model config?

yanksyoon: Sorry if I didn't understand correctly, would you mean that this should be fetched from the juju model config from default?

jdkandersson: Yes this should be done via model configurations, @weiiwang01 has a PR open for flask to do something similar

jdkandersson: https://github.com/canonical/flask-k8s-operator/pull/33

src/state.py
```diff
@@ -155,6 +156,40 @@ def _get_agent_meta_map_from_relation(
...
+    def from_charm_env(cls, env: typing.Mapping[str, str]) -> typing.Optional["ProxyConfig"]:
+        """Instantiate ProxyConfig from juju charm environment.
+
+        Args:
+            env: The charm environment variable.
+
+        Returns:
+            ProxyConfig if proxy configuration is provided, None otherwise.
+        """
+        http_proxy = env.get("HTTP_PROXY")
```
arturo-seijas, line 182: Shouldn't these be prefixed by `JUJU_CHARM_`?

src/state.py
```diff
@@ -155,6 +156,40 @@ def _get_agent_meta_map_from_relation(
...
+class ProxyConfig(BaseModel):
+    """Configuration for accessing Jenkins through proxy.
+
+    Attributes:
+        http_proxy: The http proxy URL.
+        https_proxy: The https proxy URL.
+        no_proxy: Comma separated list of hostnames to bypass proxy.
+    """
+
+    http_proxy: typing.Optional[HttpUrl]
```
arturo-seijas, line 168: Wouldn't AnyHttpUrl be more suitable?

src/state.py
```diff
@@ -155,6 +156,40 @@ def _get_agent_meta_map_from_relation(
...
+    https_proxy: typing.Optional[HttpUrl]
+    no_proxy: typing.Optional[str]
+
+    @classmethod
+    def from_charm_env(cls, env: typing.Mapping[str, str]) -> typing.Optional["ProxyConfig"]:
```
arturo-seijas, line 173: Why not directly reading the JUJU_CHARM_ proxy env vars directly?

src/state.py
```diff
@@ -155,6 +156,40 @@ def _get_agent_meta_map_from_relation(
...
+
+        Args:
+            env: The charm environment variable.
+
+        Returns:
+            ProxyConfig if proxy configuration is provided, None otherwise.
+        """
+        http_proxy = env.get("HTTP_PROXY")
+        https_proxy = env.get("HTTPS_PROXY")
+        no_proxy = env.get("NO_PROXY")
```
weiiwang01, line 184: Should we listen to `JUJU_CHARM_HTTP_PROXY`, `JUJU_CHARM_HTTPS_PROXY` and `JUJU_CHARM_NO_PROXY` instead?

src/jenkins.py
```diff
@@ -269,9 +270,22 @@ def _configure_proxy(
...
-        proxy_args += [f"'{proxy_config.username}'", f"'{proxy_config.password}'"]
+    if proxy_config.https_proxy:
+        proxy_args = [
+            f"'{proxy_config.https_proxy.host}'",
+            f"{proxy_config.https_proxy.port}",
+            f"'{proxy_config.https_proxy.user or ''}'",
+            f"'{proxy_config.https_proxy.password or ''}'",
+        ]
+    else:
+        # http proxy and https proxy value both cannot be None at the same time.
```
jdkandersson, line 281: This is confusing? What does this mean?

src/jenkins.py
```diff
@@ -297,25 +311,24 @@ def _install_plugins(
...
-    if proxy_config:
+    if proxy_config and proxy_config.http_proxy:
         command += [
-            f"-Dhttp.proxyHost={proxy_config.hostname}",
-            f"-Dhttp.proxyPort={proxy_config.port}",
-            f"-Dhttps.proxyHost={proxy_config.hostname}",
-            f"-Dhttps.proxyPort={proxy_config.port}",
+            f"-Dhttp.proxyHost={proxy_config.http_proxy.host}",
+            f"-Dhttp.proxyPort={proxy_config.http_proxy.port}",
+            f"-Dhttp.proxyUser={proxy_config.http_proxy.user or ''}",
```
jdkandersson, line 318: This looks a bit like what we did above? Any opportunities to consolidate, especially the user and password bit?

src/jenkins.py
```diff
@@ -469,18 +482,25 @@ def _get_major_minor_version(version: str) -> str:
...
+
     Returns:
         The jenkins versions from the RSS feed.
 
     Raises:
         JenkinsNetworkError: if there was an error fetching the RSS feed.
         ValidationError: if an invalid RSS feed was received.
     """
+    if proxy:
+        proxies = {"http": str(proxy.http_proxy), "https": str(proxy.https_proxy)}
```
jdkandersson, line 499: if not `str`, what are `proxy.http_proxy`?

src/state.py
```diff
@@ -159,44 +160,33 @@ class ProxyConfig(BaseModel):
...
-        hostname: The proxy server hostname.
-        port: The proxy server port.
-        username: The proxy server username for authentication.
-        password: The proxy server password for authentication.
+        http_proxy: The http proxy URL.
+        https_proxy: The https proxy URL.
         no_proxy: Comma separated list of hostnames to bypass proxy.
     """
 
-    hostname: typing.Optional[str]
```
jdkandersson: What is the reason for no longer needing these please?

src/state.py
jdkandersson: That does sound a bit better

src/state.py
jdkandersson: Oh I'm guessing these were the previous way that proxies were configured

src/state.py
jdkandersson: Yeh I think these environment variables are not quite right

src/jenkins.py
yanksyoon: Would 
```suggestion
        # http proxy and https proxy value cannot both be None since proxy_config would be parsed as None.
```
be better?

src/jenkins.py
yanksyoon: http_proxy is of type `HttpUrl` given by pydantic, so by default when user/password isn't given, it will be of type `None`. What I could suggest is to implement another method on `ProxyConfig` class to output a `typedDict` or a `NamedTuple` that can return a default value of empty string instead of `None`. Would this be better?

src/state.py
yanksyoon: Hm interesting... I've tried printing the `env` and this is what I get (irrelevant envs removed)
```
{
  "JUJU_UNIT_NAME": "jenkins-k8s/0",
  "HTTPS_PROXY": "http://192.168.0.196:3128",
  "no_proxy": "127.0.0.1,::1,localhost",
  "JUJU_VERSION": "2.9.43",
  "JUJU_CHARM_HTTP_PROXY": "",
  "NO_PROXY": "127.0.0.1,::1,localhost",
  "JUJU_CHARM_HTTPS_PROXY": "",
  "JUJU_MODEL_NAME": "stg-jenkins-k8s",
  "https_proxy": "http://192.168.0.196:3128",
  "http_proxy": "http://192.168.0.196:3128",
  "JUJU_CHARM_DIR": "/var/lib/juju/agents/unit-jenkins-k8s-0/charm",
  "JUJU_CHARM_NO_PROXY": "127.0.0.1,localhost,::1",
  "HTTP_PROXY": "http://192.168.0.196:3128",
  "PYTHONPATH": "lib:venv",
}
```
I can see that there is `http_proxy` and `https_proxy` but none of `JUJU_CHARM_HTTP_PROXY`, `JUJU_CHARM_HTTPS_PROXY`, `JUJU_CHARM_NO_PROXY`. Would this be due to my juju version?

yanksyoon: https://chat.charmhub.io/charmhub/pl/xjjptuhuxjgcffr5pg3rxkhtxw

Asked for clarification of which environment variable should be used on public mattermost!

src/jenkins.py
jdkandersson: ah yes

src/jenkins.py
jdkandersson: I guess try it and we'll see 🙂

src/state.py
yanksyoon: Fixed, thank you!

src/state.py
yanksyoon: Changed to read from `os.environ` directly, thank you!

src/jenkins.py

src/jenkins.py
yanksyoon: refactored, thank you!

src/jenkins.py
```diff
@@ -311,19 +311,20 @@ def _install_plugins(
         JenkinsPluginError: if an error occurred installing the plugin.
     """
     command = ["java"]
+    proxy_dict = proxy_config.to_str_dict() if proxy_config else {}
```
jdkandersson, line 314: hmm, why use a dict here? Why not use the data structure directly so that we get autocomplete etc?

src/state.py
yanksyoon: I remembered the reason for using `HttpUrl` over `AnyHttpUrl` writing the unit tests.
`AnyHttpUrl` wouldn't automatically determine the port from the scheme (http / https) whereas `HttpUrl` would give me 80 / 443. I need the ports to be automatically inferred and hence used the HttpUrl. Would this be okay? Otherwise I would need to write a custom parser.

src/jenkins.py
```diff
@@ -249,6 +250,36 @@ def _install_configs(container: ops.Container) -> None:
...
+    """Get proxy arguments for proxy configuration Groovy script.
+
+    Args:
+        proxy_config: The proxy settings to apply.
+
+    Yields:
+        Groovy script proxy arguments.
+    """
+    if proxy_config.https_proxy:
+        yield from (
```
jdkandersson, line 263: Perhaps just yield individual lines? That way Python doesn't need to create a tuple and then yield it out

jdkandersson (commented):
Getting close

src/jenkins.py
yanksyoon: Sure! Thanks for the comment!

jdkandersson (dismissed):

src/jenkins.py
```diff
@@ -245,30 +250,114 @@ def _install_configs(container: ops.Container) -> None:
...
+        proxy_config: The proxy settings to apply.
+
+    Yields:
+        JVM System property proxy arguments.
+    """
+    if proxy_config.http_proxy:
+        yield f"-Dhttp.proxyHost={proxy_config.http_proxy.host}",
+        yield f"-Dhttp.proxyPort={proxy_config.http_proxy.port}",
+        if proxy_config.http_proxy.user and proxy_config.http_proxy.password:
+            yield f"-Dhttp.proxyUser={proxy_config.http_proxy.user}",
```
jdkandersson, line 319: I think the trailing comma needs to be deleted

yanksyoon: Yup, just realized that, sorry about that!

jdkandersson (approved):

src/state.py
yanksyoon: Changed, thank you!

src/jenkins.py
yanksyoon: Changed, thank you!

src/state.py
arturo-seijas: Sure. Thanks for clarifying

src/jenkins.py
```diff
@@ -245,30 +250,114 @@ def _install_configs(container: ops.Container) -> None:
...
+            yield f"-Dhttp.proxyPassword={proxy_config.http_proxy.password}"
+    if proxy_config.https_proxy:
+        yield f"-Dhttps.proxyHost={proxy_config.https_proxy.host}"
+        yield f"-Dhttps.proxyPort={proxy_config.https_proxy.port}"
+        if proxy_config.https_proxy.user and proxy_config.https_proxy.password:
+            yield f"-Dhttps.proxyUser={proxy_config.https_proxy.user}"
+            yield f"-Dhttps.proxyPassword={proxy_config.https_proxy.password}"
+    if proxy_config.no_proxy:
+        formatted_no_proxy_hosts = "|".join(proxy_config.no_proxy.split(","))
+        yield f'-Dhttp.nonProxyHosts="{formatted_no_proxy_hosts}"'
```
amandahla, line 329: minor/can-be-addressed-later: Does it exists a https.nonProxyHosts configuration?

amandahla (approved):
Nice :) Just left a question.

src/jenkins.py
yanksyoon: Oh, the comment got removed during the refactor! I will add it back in a future push!
Http and https shares the same configuration for no proxy! :D

weiiwang01 (approved):

src/state.py
```diff
@@ -155,6 +156,37 @@ def _get_agent_meta_map_from_relation(
...
+        """
+        http_proxy = os.environ.get("JUJU_CHARM_HTTP_PROXY")
+        https_proxy = os.environ.get("JUJU_CHARM_HTTPS_PROXY")
+        no_proxy = os.environ.get("JUJU_CHARM_NO_PROXY")
+        if not http_proxy and not https_proxy:
+            return None
+        # Mypy doesn't understand str is supposed to be converted to HttpUrl by Pydantic.
+        return cls(
+            http_proxy=http_proxy, https_proxy=https_proxy, no_proxy=no_proxy  # type: ignore
+        )
```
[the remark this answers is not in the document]
yanksyoon: That's a great tip! Thank you for letting me know!

PR #23, opened 2023-08-15
feat: plugins configuration

yanksyoon:
This PR:
- changes `update-time-range` configuration option to `restart-time-range` which handles administrative tasks such as removing unauthorized plugins and updating jenkins when a new patch version is available.
- refactors code styles to import directly from the `ops` library.
- refactors code due to many of the code not passing `too-complex` lint.

Checklist:
- [x] Unit tests passing
- [x]  Integration tests passing
- [x] Documentation updated


config.yaml
```diff
@@ -2,10 +2,20 @@
...
+  restart-time-range:
     type: string
     description: >
-      Preferred UTC time range for LTS patch updates in 24 hour  format. If empty, the update will
-      take place whenever an update is available. For example, 03-05 will allow updates to take
-      place from 3AM UTC to 5AM UTC. Awaits for running job completion for 5 minutes. 
+      Preferred UTC time range for restarting Jenkins in 24 hour format. If empty, restart will
+      take place whenever Jenkins needs to restart. Jenkins will to restart on the following
+      occasions. 1. Patched Jenkins version within the same LTS available. 2. Plugins that are not
+      part of `plugins` configuration option is detected.
```
arturo-seijas, line 11: s/is/are

config.yaml
```diff
@@ -2,10 +2,20 @@
...
 
 options:
-  update-time-range:
+  restart-time-range:
     type: string
     description: >
-      Preferred UTC time range for LTS patch updates in 24 hour  format. If empty, the update will
-      take place whenever an update is available. For example, 03-05 will allow updates to take
-      place from 3AM UTC to 5AM UTC. Awaits for running job completion for 5 minutes. 
+      Preferred UTC time range for restarting Jenkins in 24 hour format. If empty, restart will
```
arturo-seijas, line 8: This looks simpler to me:  `Preferred UTC time rangein 24 hour format  for restarting Jenkins. If empty, restart will take place if:`

config.yaml
```diff
@@ -2,10 +2,20 @@
...
+      take place whenever Jenkins needs to restart. Jenkins will to restart on the following
+      occasions. 1. Patched Jenkins version within the same LTS available. 2. Plugins that are not
+      part of `plugins` configuration option is detected.
+      For example, 03-05 will allow Jenkins restart to take place from 3AM UTC to 5AM UTC. 
+      Awaits for running job completion for 5 minutes.
     default: ""
+  plugins:
+    type: string
+    description: >
+      List of allowed plugin short names separated by commas. Leaving this empty will allow any
```
arturo-seijas, line 18: This seems simpler to me
```
Comma-separated list of allowed plugin short names. If empty, any plugin can be installed.  Plugins installed by the user and their dependencies will be removed automatically if not on the list. Included plugins are not automatically installed.

File: docs/how-to/configure-restart-time-range.md [hunk 4]
```diff
@@ -1,13 +1,18 @@
...
+# How to configure restart-time-range
 
-### Configure update-time-range
+### Configure restart-time-range
+
+Use the `restart-time-range` configuration to set the time interval when `jenkins-k8s`
+
+1. automatically applies the latest patches for the current
+   [LTS](https://www.jenkins.io/download/lts/) version.
+2. automatically remove plugins that are installed on the system but is not part of the `plugins`
```
arturo-seijas, line 9: s/but is/but are

metadata.yaml
```diff
@@ -7,7 +7,7 @@ assumes:
 display-name: Jenkins K8s
 summary: Jenkins Continuous Integration Server
 maintainers:
-  - IS DevOps <launchpad.net/~canonical-is-devops>
+  - https://launchpad.net/~canonical-is-devops
```
arturo-seijas, line 10: Next time, I suggest you open a separate PR for this kind of changes, so that PRs are move atomic and can be easily reverted if needed

src-docs/jenkins.py.md
```diff
@@ -340,6 +346,39 @@ Infer agent name from unit name.
...
+ 
+ - <b>`plugins`</b>:  The list of plugins that can be installed. 
+ - <b>`container`</b>:  The workload container. 
+ - <b>`client`</b>:  The Jenkins API client. 
+
+
+
+**Raises:**
+ 
+ - <b>`JenkinsPluginError`</b>:  if there was an error removing unlisted plugin.
```
arturo-seijas, line 377: /removing unlisted/removing an unlisted

src-docs/status.py.md
```diff
@@ -0,0 +1,32 @@
...
+
+<a href="../src/status.py#L11"><img align="right" style="float:right;" src="https://img.shields.io/badge/-source-cccccc?style=flat-square"></a>
+
+## <kbd>function</kbd> `get_priority_status`
+
+```python
+get_priority_status(statuses: Iterable[StatusBase]) → StatusBase
+```
+
+Get status to display out of all possible statuses returned by charm components.
```
arturo-seijas, line 19: This wording seems a bit confusing to me

src/charm.py
```diff
@@ -113,85 +106,125 @@ def _on_jenkins_pebble_ready(self, event: PebbleReadyEvent) -> None:
...
             event.defer()
             return
-        credentials = jenkins.get_admin_credentials(self._jenkins_container)
+        credentials = jenkins.get_admin_credentials(container)
         event.set_results({"password": credentials.password})
 
-    def _on_update_status(self, _: UpdateStatusEvent) -> None:
-        """Handle update status event.
+    def _remove_unlisted_plugins(self, container: ops.Container) -> ops.StatusBase:
+        """Remove plugins that are installed but not allowed.
```
arturo-seijas, line 147: s/plugins that are installed but not allowed/installed but not allowed plugins
The simpler the grammar, the easier to read :)

src/charm.py
```diff
@@ -113,85 +106,125 @@ def _on_jenkins_pebble_ready(self, event: PebbleReadyEvent) -> None:
...
 
         self.unit.set_workload_version(latest_patch_version)
-        self.unit.status = ActiveStatus()
+        return ops.ActiveStatus()
+
+    def _on_update_status(self, _: ops.UpdateStatusEvent) -> None:
+        """Handle update status event.
+
+        On Update status:
+        1. Remove plugins that are installed but are not allowed by plugins config value.
```
arturo-seijas, line 211: s/plugins that are installed but/installed plugins that

src/jenkins.py
```diff
@@ -705,3 +708,196 @@ def get_agent_name(unit_name: str) -> str:
...
+    config["jenkins"]["systemMessage"] = message
+    container.push(
+        JCASC_CONFIG_FILE_PATH, yaml.dump(config), encoding="utf-8", user=USER, group=GROUP
+    )
+
+
+def _traverse_dependencies(
+    plugin: str, dependency_lookup: typing.Mapping[str, typing.Iterable[str]], seen: set[str]
+) -> typing.Iterable[str]:
+    """Traverse through plugin dependencies that have not yet been traversed recursively.
```
arturo-seijas, line 788: "Recursively traverse through plugin dependencies" instead?

src/jenkins.py
```diff
@@ -705,3 +708,196 @@ def get_agent_name(unit_name: str) -> str:
...
+
+    Args:
+        plugin: The plugin to recurse through the dependencies.
+        dependency_lookup: The plugin and it's dependencies lookup table.
```
arturo-seijas, line 792: s/it's/its

src/jenkins.py
```diff
@@ -705,3 +708,196 @@ def get_agent_name(unit_name: str) -> str:
...
+        seen: a set indicating whether a plugin has already been traversed.
```
arturo-seijas, line 793: s/traversed/visited? I think that's the usual terminology used in search algorithms

src/jenkins.py
```diff
@@ -705,3 +708,196 @@ def get_agent_name(unit_name: str) -> str:
...
+    seen.add(plugin)
+    for dependency in dependency_lookup.get(plugin, []):
+        yield from _traverse_dependencies(dependency, dependency_lookup, seen)
+
+
+def _get_allowed_plugins(
+    top_level_plugins: typing.Iterable[str],
+    dependency_lookup: typing.Mapping[str, typing.Iterable[str]],
+) -> typing.Iterable[str]:
+    """Get the plugin short names of allowed plugins and its dependencies.
```
arturo-seijas, line 810: s/its/their

src/jenkins.py
```diff
@@ -705,3 +708,196 @@ def get_agent_name(unit_name: str) -> str:
...
+
+    Args:
+        top_level_plugins: The allowed plugins short names to add to allowed plugins with its
```
arturo-seijas, line 813: s/its/their

src/jenkins.py
```diff
@@ -705,3 +708,196 @@ def get_agent_name(unit_name: str) -> str:
...
+            dependencies.
+        dependency_lookup: The plugin dependency lookup table.
+
+    Yields:
+        The allowed plugin short name.
+    """
+    seen: set[str] = set()
+    for plugin in top_level_plugins:
+        if plugin in seen:
```
arturo-seijas, line 822: Why not
```
if not plugin in seen:
        yield from _traverse_dependencies(plugin, dependency_lookup, seen)
```
 ?

src/jenkins.py
```diff
@@ -705,3 +708,196 @@ def get_agent_name(unit_name: str) -> str:
...
+            continue
+        yield from _traverse_dependencies(plugin, dependency_lookup, seen)
+
+
+def _get_top_level_plugins(
+    plugins: typing.Iterable[str], dependency_lookup: typing.Mapping[str, typing.Iterable[str]]
+) -> typing.Iterable[str]:
+    """Get top level plugins that are not dependencies to other plugins.
```
arturo-seijas, line 830: s/to/of?

src/jenkins.py
```diff
@@ -705,3 +708,196 @@ def get_agent_name(unit_name: str) -> str:
...
+
+    Args:
+        plugins: Plugins to extract top level plugins from.
+        dependency_lookup: The dependency lookup table.
+
+    Returns:
+        All plugins that are not dependency to another plugin.
```
arturo-seijas, line 837: Same here?

src/jenkins.py
```diff
@@ -705,3 +708,196 @@ def get_agent_name(unit_name: str) -> str:
...
+
+    return set(plugins) - dependent_plugins
+
+
+def remove_unlisted_plugins(
+    plugins: typing.Iterable[str] | None,
+    container: ops.Container,
+    client: jenkinsapi.jenkins.Jenkins | None = None,
+) -> None:
+    """Remove plugins that are not a part of list of desired plugins.
```
arturo-seijas, line 851: s/a part of/in the

src/jenkins.py
```diff
@@ -705,3 +708,196 @@ def get_agent_name(unit_name: str) -> str:
...
+        The dependency lookup table.
+    """
+    dependency_lookup: dict[str, tuple[str, ...]] = {}
+    for line in plugin_dependency_outputs:
+        match = re.match(PLUGIN_LINE_CAPTURE, line)
+        if not match:
+            continue
+        plugin, dependencies = match.group(1), match.group(3)
+        if not dependencies:
+            dependency_lookup[plugin] = ()
```
arturo-seijas, line 757: If using a defaultdict, you can remove this line

src/jenkins.py
```diff
@@ -705,3 +708,196 @@ def get_agent_name(unit_name: str) -> str:
...
+) -> typing.Iterable[str]:
+    """Traverse through plugin dependencies that have not yet been traversed recursively.
+
+    Args:
+        plugin: The plugin to recurse through the dependencies.
+        dependency_lookup: The plugin and it's dependencies lookup table.
+        seen: a set indicating whether a plugin has already been traversed.
+
+    Yields:
+        Plugin and it's dependents.
```
arturo-seijas, line 796: s/it's/its

src/jenkins.py
```diff
@@ -705,3 +708,196 @@ def get_agent_name(unit_name: str) -> str:
...
+    """
+    if plugin in seen or plugin not in dependency_lookup:
+        return
+    yield plugin
+    seen.add(plugin)
+    for dependency in dependency_lookup.get(plugin, []):
```
arturo-seijas, line 802: Do you need the default value?

src/jenkins.py
```diff
@@ -705,3 +708,196 @@ def get_agent_name(unit_name: str) -> str:
...
+        All plugins that are not dependency to another plugin.
+    """
+    dependent_plugins: set[str] = set()
+    for _, dependencies in dependency_lookup.items():
+        dependent_plugins = dependent_plugins.union(dependencies)
+
+    return set(plugins) - dependent_plugins
+
+
+def remove_unlisted_plugins(
```
arturo-seijas, line 846: I'm not seeing any call where the `client` argument is set. Perhaps you'd want to remove it

tests/unit/test_state.py
```diff
@@ -129,3 +129,29 @@ def test_proxyconfig_from_charm_env(
     assert config.http_proxy == proxy_config.http_proxy
     assert config.https_proxy == proxy_config.https_proxy
     assert config.no_proxy == proxy_config.no_proxy
+
+
+def test_plugins_config_none(harness: Harness):
```
arturo-seijas, line 134: You should try to avoid using harness to test any piece of code different from the charm itself and the libraries. Perhaps you can define and instantiate a very simple charm

src/jenkins.py
```diff
@@ -705,3 +708,196 @@ def get_agent_name(unit_name: str) -> str:
...
+        match = re.match(PLUGIN_LINE_CAPTURE, line)
+        if not match:
+            continue
+        plugin, dependencies = match.group(1), match.group(3)
+        if not dependencies:
+            dependency_lookup[plugin] = ()
+            continue
+        try:
+            dependency_lookup[plugin] = tuple(
+                _get_plugin_name(dependency) for dependency in dependencies.split(", ")
```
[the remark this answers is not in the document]
yanksyoon: Yup, there is always a comma after the whitespace.
An example of the output of the script is:
```
instance-identity (173.va_37c494ec4e5) => [bouncycastle-api (2.27)]
javax-mail-api (1.6.2-8) => [javax-activation-api (1.2.0-4)]
javax-activation-api (1.2.0-5) => []
structs (324.va_f5d6774f3a_d) => [javax-activation-api (1.2.0-2) optional, javax-mail-api (1.6.2-5) optional, instance-identity (3.1) optional]
script-security (1264.vecf66020eb_7d) => [caffeine-api (2.9.3-65.v6a_47d0f4d1fe)]
configuration-as-code (1625.v27444588cc3d) => [caffeine-api (2.9.3-65.v6a_47d0f4d1fe), commons-text-api (1.10.0-36.vc008c8fcda_7b_), snakeyaml-api (1.33-95.va_b_a_e3e47b_fa_4)]
commons-lang3-api (3.13.0-62.v7d18e55f51e2) => []
bouncycastle-api (2.29) => []
caffeine-api (3.1.8-133.v17b_1ff2e0599) => []
snakeyaml-api (1.33-95.va_b_a_e3e47b_fa_4) => []
Result: [Plugin:commons-text-api, Plugin:instance-identity, Plugin:javax-mail-api, Plugin:javax-activation-api, Plugin:structs, Plugin:script-security, Plugin:configuration-as-code, Plugin:commons-lang3-api, Plugin:bouncycastle-api, Plugin:caffeine-api, Plugin:snakeyaml-api]
```

tests/unit/test_status.py
```diff
@@ -0,0 +1,74 @@
...
+                expected_status := ops.ActiveStatus("I have a message"),  # type: ignore
+                ops.ActiveStatus("I have a message too"),
+            ],
+            expected_status,
+            id="same statuses with messages",
+        ),
+    ],
+)
+# pylint: enable=unused-variable,undefined-variable,too-many-locals
+def test__get_priority_status(
```
arturo-seijas, line 63: There are two underscores here

src/jenkins.py
yanksyoon: I usually do the not continue case since I prefer the code not nested. I don't mind changing such simple cases though. Updating!

src/jenkins.py
yanksyoon: I'm not sure I fully understand what you mean,
the client is initialized in the following code
`client = client if client is not None else _get_client(get_admin_credentials(container))`
and is used to run groovy script, deleting plugins and restarting jenkins.
Would this be what you meant?

src/jenkins.py
arturo-seijas: Up to you. You don't have to change it

src/jenkins.py
arturo-seijas: No, what I mean is that the function signature defined a `client` argument that isn't passed in any function call

tests/unit/test_state.py
yanksyoon: Since the charm (`ops.CharmBase`) requires a `ops.Framework`, I think a harness would need to be injected anyways. Would there be an approach that I'm not seeing that's simpler?
This is the [example](https://github.com/canonical/loki-k8s-operator/blob/0daf783317b32fe31f6a4500ff01f98c406aed49/tests/unit/test_consumer.py#L18) I found in the loki-k8s codebase.

Otherwise, I could mock the framework which would involve a bit of defining the return types.
```
class TestingCharm(ops.CharmBase):
        """A simple charm for testing."""

    mock_framework = MagicMock(spec=ops.Framework)
    # mock_framework requires defining some return values to avoid errors like the following:
    # Invalid time range <MagicMock name='mock.model.config.get()' id='139838944989616'>, time range must be 
    # an integer.
    config = state.State.from_charm(TestingCharm(mock_framework))
```

src/jenkins.py
yanksyoon: I need to have the keys to build the set of all plugins, used by `plugins_lookup.keys()` later to subtract installed plugins set from all plugins set. :)

metadata.yaml
yanksyoon: Yup, I should have opened a new branch! Thanks for the reminder!

src/charm.py
yanksyoon: Definately, thank you for the suggestion!

src/jenkins.py
yanksyoon: Removed, with improvement from defaultdict, thank you for the comment!

src/jenkins.py
yanksyoon: Oh, the client is initialized inside the code but for test cases, the client is initialized outside as a mock and is injected. Hence the client in the parameters. :)

tests/unit/test_status.py
yanksyoon: Nice catch! Missed it from the refactoring!

yanksyoon: Thanks for all the comments above, resolved them :D

tests/unit/test_state.py
arturo-seijas: How about https://github.com/canonical/saml-integrator-operator/blob/main/tests/unit/test_charm_state.py#L33?

src/jenkins.py
arturo-seijas: The defaultdict can return an empty set by default, iirc. Am I wrong?

src/jenkins.py
arturo-seijas: Wouldn't it lead to a simpler code to extract that client into a property and mock it? It ill also simplify the logic in L686

src/jenkins.py
yanksyoon: It can, however, I still need the keys to be registered on the lookup table later on even if they are by default empty. For example, if I have
`{"a": [], "b": [], "c": []}`, the default dict implementation would be fine with `{}`. However, the keys information is lost. I later need the `plugins_lookup.keys()` which with default dict, would return `[]` if I don't register the keys.

tests/unit/test_state.py
yanksyoon: Oof why did I think of mocking the framework but not the charm? 😵‍💫 Updated, thank you!

src/jenkins.py
arturo-seijas: Noted ;)

tests/unit/test_state.py
```diff
@@ -36,12 +38,12 @@ def test_no_time_range_config(time_range: str, harness: Harness):
     act: when state is instantiated.
     assert: state without time range is returned.
     """
-    harness.update_config({"update-time-range": time_range})
+    harness.update_config({"restart-time-range": time_range})
```
arturo-seijas, line 41: I think you can get rid of the harness here too

tests/unit/test_state.py
```diff
@@ -17,7 +19,7 @@ def test_state_invalid_time_config(harness: Harness):
     act: when state is initialized through from_charm method.
     assert: CharmConfigInvalidError is raised.
     """
-    harness.update_config({"update-time-range": "-1"})
+    harness.update_config({"restart-time-range": "-1"})
```
arturo-seijas, line 22: I think you can get rid of the harness here too

tests/unit/test_state.py
```diff
@@ -129,3 +131,30 @@ def test_proxyconfig_from_charm_env(
...
+    assert config.plugins is None
+
+
+def test_plugins_config(harness: Harness):
+    """
+    arrange: given a charm with comma separated plugins.
+    act: when state is initialized from charm.
+    assert: plugin state contains an iterable of plugins.
+    """
+    harness.update_config({"plugins": "hello, world"})
```
arturo-seijas, line 155: I think you can get rid of the harness here too

yanksyoon: Updated, thank you!

arturo-seijas (dismissed):
LGTM

jdkandersson:
The description says it is still WIP?

yanksyoon:
@jdkandersson woops, i've only edited the title, updated!

src/charm.py
```diff
@@ -113,85 +106,125 @@ def _on_jenkins_pebble_ready(self, event: PebbleReadyEvent) -> None:
...
-            return
+        original_status = self.unit.status.name
+        try:
+            jenkins.remove_unlisted_plugins(plugins=self.state.plugins, container=container)
+        except (jenkins.JenkinsPluginError, jenkins.JenkinsError) as exc:
+            logger.error("Failed to remove unlisted plugin, %s", exc)
+            return ops.StatusBase.from_name(original_status, "Failed to remove unlisted plugin.")
+        except TimeoutError as exc:
+            logger.error("Failed to restart jenkins after removing plugin, %s", exc)
+            return ops.BlockedStatus("Failed to restart Jenkins (remove plugin)")
```
jdkandersson, line 163: The `(remove plugin)` could be confusing, is the intent to indicate why the restart was triggered?

src/charm.py
```diff
@@ -113,85 +106,125 @@ def _on_jenkins_pebble_ready(self, event: PebbleReadyEvent) -> None:
...
+        Args:
+            container: The Jenkins workload container.
+
+        Returns:
+            The unit status of the charm after the operation.
+        """
+        original_status = self.unit.status.name
+        self.unit.status = ops.StatusBase.from_name(original_status, "Checking for updates.")
         try:
             latest_patch_version = jenkins.get_updatable_version(proxy=self.state.proxy_config)
```
jdkandersson, line 181: Did we discuss not returning the version and instead indicating whether Jenkins needs to be updated?

src/charm.py
```diff
@@ -113,85 +106,125 @@ def _on_jenkins_pebble_ready(self, event: PebbleReadyEvent) -> None:
...
+
+        On Update status:
+        1. Remove plugins that are installed but are not allowed by plugins config value.
+        2. Update Jenkins patch version if available and is within restart-time-range config value.
+        """
+        container = self.unit.get_container(self.state.jenkins_service_name)
+        if not container.can_connect():
+            return
+
+        if self.state.restart_time_range and not self.state.restart_time_range.check_now():
```
jdkandersson, line 218: hmm, I'm slightly dissatisfied with the state exposing the `check_now` function, the state should be more like a data object rather than providing functionality. Could this just be in a separate function which takes the state as an argument and returns whether it is ok to restart?

src/jenkins.py
```diff
@@ -705,3 +709,195 @@ def get_agent_name(unit_name: str) -> str:
...
+    """Build a lookup table of plugin short name to list of dependency plugin's short names.
+
+    Args:
+        plugin_dependency_outputs: The plugin dependency output from Jenkins Groovy script.
+
+    Returns:
+        The dependency lookup table.
+    """
+    # coverage tool thinks defaultdict() has an exit condition.
+    dependency_lookup: dict[str, tuple[str, ...]] = defaultdict(lambda: ())  # pragma: no cover
```
jdkandersson, line 752: I think you can do `defaultdict(tuple)`

src/jenkins.py
```diff
@@ -705,3 +709,195 @@ def get_agent_name(unit_name: str) -> str:
...
+                _get_plugin_name(dependency) for dependency in dependencies.split(", ")
+            )
+        except ValidationError as exc:
+            logger.error("Invalid plugin dependency, %s", exc)
+            continue
+
+    return dependency_lookup
+
+
+def _set_jenkins_system_message(message: str, container: ops.Container) -> None:
```
jdkandersson, line 771: Do you think we can do a bit more to order these functions logically? I'm surprised this function is listed straight after the `_build_dependencies_lookup` function which hasn't been used yet?

src/jenkins.py
```diff
@@ -705,3 +709,195 @@ def get_agent_name(unit_name: str) -> str:
...
+
+    Yields:
+        Plugin and its dependents.
+    """
+    if plugin in seen or plugin not in dependency_lookup:
+        return
+    yield plugin
+    seen.add(plugin)
+    for dependency in dependency_lookup[plugin]:
+        yield from _traverse_dependencies(dependency, dependency_lookup, seen)
```
jdkandersson, line 804: What about `yield from itertools.chain(_traverse_dependencies(dependency, dependency_lookup, seen) for dependency in dependency_lookup[plugin])`

More info: https://docs.python.org/3/library/itertools.html#itertools.chain.from_iterable

src/jenkins.py
```diff
@@ -705,3 +709,195 @@ def get_agent_name(unit_name: str) -> str:
...
+
+
+def _get_allowed_plugins(
```
jdkandersson, line 807: Hmm, can you think of how to now need this method and `_traverse_dependencies`? E.g., default `seen` to `None` and create a new set in that case

src/jenkins.py
```diff
@@ -705,3 +709,195 @@ def get_agent_name(unit_name: str) -> str:
...
+
+    Raises:
+        JenkinsPluginError: if there was an error removing unlisted plugin.
+        JenkinsError: if there was an error restarting Jenkins after removing the plugin.
+        TimeoutError: if it took too long to restart Jenkins after removing the plugin.
+    """
+    if not plugins:
+        return
+
+    client = client if client is not None else _get_client(get_admin_credentials(container))
```
jdkandersson, line 866: This is a bit weird, when would the client be `None` and not?

src/jenkins.py
```diff
@@ -705,3 +709,195 @@ def get_agent_name(unit_name: str) -> str:
...
+    Args:
+        top_level_plugins: The allowed plugins short names to add to allowed plugins with their
+            dependencies.
+        dependency_lookup: The plugin dependency lookup table.
+
+    Yields:
+        The allowed plugin short name.
+    """
+    seen: set[str] = set()
+    for plugin in top_level_plugins:
```
jdkandersson, line 822: Shouldn't this be something like `plugin in allowed_plugins`?

src/jenkins.py
```diff
@@ -705,3 +709,195 @@ def get_agent_name(unit_name: str) -> str:
...
+            yield from _traverse_dependencies(plugin, dependency_lookup, seen)
+
+
+def _get_top_level_plugins(
+    plugins: typing.Iterable[str], dependency_lookup: typing.Mapping[str, typing.Iterable[str]]
+) -> typing.Iterable[str]:
+    """Get top level plugins that are not dependencies of other plugins.
+
+    Args:
+        plugins: Plugins to extract top level plugins from.
```
jdkandersson, line 833: Why is this argument needed? Can't this be built entirely based on `dependency_lookup`?

src/jenkins.py
```diff
@@ -705,3 +709,195 @@ def get_agent_name(unit_name: str) -> str:
...
+    try:
+        client.delete_plugins(plugin_list=plugins_to_remove, restart=False)
+    except jenkinsapi.custom_exceptions.JenkinsAPIException as exc:
+        logger.error("Failed to remove the following plugins: %s, %s", plugins_to_remove, exc)
+        raise JenkinsPluginError("Failed to remove plugins.") from exc
+    logger.debug("Removed %s", plugins_to_remove)
+
+    top_level_plugins = _get_top_level_plugins(plugins_to_remove, dependency_lookup)
+    _set_jenkins_system_message(
+        message="The following plugins have been removed by the system administrator: "
```
jdkandersson, line 892: Perhaps include the reason why they have been removed?

src/status.py
```diff
@@ -0,0 +1,42 @@
...
+    """Get status to display out of all possible statuses returned by charm components.
+
+    Args:
+        statuses: Statuses returned by components of the charm.
+
+    Returns:
+        The final status to display.
+    """
+
+    def get_status_priority(status: ops.StatusBase) -> int:
```
jdkandersson, line 21: What is the reason for this function to be embedded in the `get_priority_status` function?

src/status.py
```diff
@@ -0,0 +1,42 @@
...
+        """
+        priority = {
+            ops.ErrorStatus.name: 0,
+            ops.BlockedStatus.name: 2,
+            ops.MaintenanceStatus.name: 4,
+            ops.WaitingStatus.name: 6,
+            ops.ActiveStatus.name: 8,
+        }.get(status.name)
+        priority = typing.cast(int, priority)
+        if status.message:
```
jdkandersson, line 38: I don't follow this?

src/status.py
```diff
@@ -0,0 +1,42 @@
...
+            return priority - 1
+        return priority
+
+    return sorted(statuses, key=get_status_priority)[0]
```
jdkandersson, line 42: What about `return sorted(statuses, key= priority.get)`

jdkandersson (commented):
Haven't reviewed the tests yet

src/jenkins.py
jdkandersson: And, if we're not using the default diction functionality, why use it please?

src/status.py
jdkandersson: Oh I see, it is so that if there is a status message, that one gets picked

jdkandersson: Isn't there a then by type of functionality in python?

jdkandersson: Perhaps something like this: https://stackoverflow.com/questions/5944630/sort-a-list-of-dicts-by-x-then-by-y

config.yaml
```diff
@@ -2,10 +2,19 @@
...
+      take place whenever Jenkins needs to restart. Jenkins will to restart on the following
+      occasions. 1. Patched Jenkins version within the same LTS is available. 2. Plugins that are
+      not part of `plugins` configuration option are detected.
+      For example, 03-05 will allow Jenkins restart to take place from 3AM UTC to 5AM UTC. 
+      Awaits for running job completion for 5 minutes.
     default: ""
+  plugins:
+    type: string
+    description: >
+      Comma-separated list of allowed plugin short names. If empty, any plugin can be installed.
```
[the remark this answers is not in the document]
yanksyoon: I'm afraid that will be more than 80+ which might result in a very long list of defaults. Perhaps this is better suited when the plugin integration tests are done, when I have more information on whether it is a user-installable plugin vs a dependency plugin.
For example, `git` is a user-installable plugin as well as a dependency plugin to other plugins. I can do a bit more data scraping to find this out.

docs/how-to/configure-installable-plugins.md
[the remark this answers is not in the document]
yanksyoon: Raised in another PR! https://github.com/canonical/jenkins-k8s-operator/pull/24

src/charm.py
yanksyoon: Yes, the two jenkins "management" functions `_remove_unlisted_plugins` and `_update_jenkins_version` both may block on failure to restart, so I thought such status message could help differentiate the two.
Perhaps "Failed to restart Jenkins after removing plugins" might be better?

src/charm.py
yanksyoon: Yup, I think it is better suited on a different refactor branch! I've got it separately marked as #TODO comment locally :)

src/jenkins.py
yanksyoon: I think it is *implicitly* getting used when the code traverses the `dependency_lookup` recursively. It can help be defensive about the returned value from Jenkins, although I trust the jenkins output.
For example: 
L803: `for dependency in dependency_lookup[plugin]:` might throw a `KeyError` on a regular dict, but with `defaultdict` it helps just skip the traversal of plugins that are not found.

src/jenkins.py
yanksyoon: The client is `None` when the charm calls it, the client is `MagicMock(spec=jenkinsapi.Jenkins)` in unit tests :D, this is only here to facilitate the tests.

src/jenkins.py
yanksyoon: This is a list of plugins that represents a set all plugins (plugins + dependencies) and it is used to find out which plugin is a top level plugin `(plugin + dependencies) - dependencies` that should be present when showing Jenkins system message. I'm not sure I could come up with a better name for this variable. Perhaps `plugin_and_dependencies`?

src/jenkins.py
yanksyoon: I think the line after next "To allow the plugins, please include them in the plugins configuration of the charm." is quite explanatory. Shall I move this message up?

src/jenkins.py
yanksyoon: This will be updated in a follow up refactor PR.

src/jenkins.py
yanksyoon: renamed :)

src/jenkins.py
yanksyoon: Renamed the function. Thank you :)

src/status.py
yanksyoon: The key sorting function needs to prioritise any status with message first, this function has been refactored to cater for that with a lambda function.

src/status.py
yanksyoon: Refactored, thank you!

src/status.py
yanksyoon: This function has been refactored as per the suggestion, thank you!

src/jenkins.py
yanksyoon: They have been merged together, thank you!

src/jenkins.py
yanksyoon: This function has been refactored, in my opinion it is better to keep the for-loop in the refactored code to make it readable.

src/jenkins.py
yanksyoon: Reordered by the order in which they are called. Thank you!

src/jenkins.py
yanksyoon: Refactored to use normal dict and handle KeyError with warnings

src/charm.py
yanksyoon: This has been updated in a separate module, thank you for the suggestion!

config.yaml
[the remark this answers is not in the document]
yanksyoon: This is discussed to be added in the integration test one by one. Thank you for the suggestion!

config.yaml
```diff
@@ -2,10 +2,19 @@
...
-      take place whenever an update is available. For example, 03-05 will allow updates to take
-      place from 3AM UTC to 5AM UTC. Awaits for running job completion for 5 minutes. 
+      Preferred UTC time range in 24 hour format for restarting Jenkins. If empty, restart will
+      take place whenever Jenkins needs to restart. Jenkins will to restart on the following
+      occasions. 1. Patched Jenkins version within the same LTS is available. 2. Plugins that are
+      not part of `plugins` configuration option are detected.
+      For example, 03-05 will allow Jenkins restart to take place from 3AM UTC to 5AM UTC. 
+      Awaits for running job completion for 5 minutes.
     default: ""
+  plugins:
```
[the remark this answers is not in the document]
yanksyoon: Changed, thank you for the suggestion!

config.yaml
```diff
@@ -2,10 +2,19 @@
...
 options:
-  update-time-range:
+  restart-time-range:
     type: string
     description: >
-      Preferred UTC time range for LTS patch updates in 24 hour  format. If empty, the update will
-      take place whenever an update is available. For example, 03-05 will allow updates to take
-      place from 3AM UTC to 5AM UTC. Awaits for running job completion for 5 minutes. 
+      Preferred UTC time range in 24 hour format for restarting Jenkins. If empty, restart will
+      take place whenever Jenkins needs to restart. Jenkins will to restart on the following
```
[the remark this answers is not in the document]
yanksyoon: Nice catch! Thank you!

src/jenkins.py
```diff
@@ -768,69 +766,46 @@ def _build_dependencies_lookup(
...
     Yields:
         The allowed plugin short name.
     """
-    seen: set[str] = set()
-    for plugin in top_level_plugins:
-        if plugin not in seen:
-            yield from _traverse_dependencies(plugin, dependency_lookup, seen)
+    # mypy doesn't understand that we can reassign the type and it cannot be None aftwards.
+    if seen is None:
+        seen: set[str] = set()  # type: ignore
```
jdkandersson, line 787: this should be solved by removing `: set[str]`

src/jenkins.py
```diff
@@ -843,6 +818,21 @@ def _get_top_level_plugins(
...
+def _set_jenkins_system_message(message: str, container: ops.Container) -> None:
+    """Set a system message on Jenkins.
+
+    Args:
+        message: The system message to display.
+        container: The Jenkins workload container.
+    """
+    jcasc_yaml = container.pull(JCASC_CONFIG_FILE_PATH, encoding="utf-8").read()
+    config = yaml.safe_load(jcasc_yaml)
+    config["jenkins"]["systemMessage"] = message
```
jdkandersson, line 830: What about `KeyError`s that could happen?

src/status.py
```diff
@@ -17,26 +25,6 @@ def get_priority_status(statuses: typing.Iterable[ops.StatusBase]) -> ops.Status
...
-            ops.WaitingStatus.name: 6,
-            ops.ActiveStatus.name: 8,
-        }.get(status.name)
-        priority = typing.cast(int, priority)
-        if status.message:
-            return priority - 1
-        return priority
-
-    return sorted(statuses, key=get_status_priority)[0]
+    return sorted(statuses, key=lambda item: (PRIORITY_MAP[item.name] - int(bool(item.message))))[
```
jdkandersson, line 28: Close, let's change to lambda returning `(PRIORITY_MAP[item.name], int(bool(item.message)))`, check whether this will get the correct order, perhaps the second element needs a minus

tests/unit/test_jenkins.py
```diff
@@ -1039,59 +1039,6 @@ def test__build_dependencies_lookup(
...
-    assert: the plugin and its dependencies are traversed.
-    """
-    traversed_plugins = jenkins._traverse_dependencies(plugin, plugins_lookup, set())
-
-    assert tuple(traversed_plugins) == expected_plugins
-
-
 @pytest.mark.parametrize(
     "top_level_plugins, plugins_lookup, expected_allowed_plugins",
     [
```
jdkandersson, line 1044: For test cases, I would suggest ordering them logically. For example:

- all empty
- one plugin in top level and
  - nothing in lookup
  - one in lookup but its different
  - one in lookup and it matches the plugin with
    - no extra dependencies
    - one dependency and
      - no nested depedency
      - single nested dependency
    - multiple dependencies
- two plugins where
  - none of them have dependencies
  - one of them has and the other doesn't have dependencies (2 test cases here)
  - both of them have a single dependency and
    - the dependencies are different
    - they are the same
  - both of them have multiple dependencies and
    - there is no overlap
    - one is the same (2 test cases here)
    - both are the same

Let's not use a set so that we can check whether the seen plugins check works

src/jenkins.py
```diff
@@ -705,3 +708,186 @@ def get_agent_name(unit_name: str) -> str:
...
+    # mypy doesn't understand that we can reassign the type and it cannot be None aftwards.
+    if seen is None:
+        seen: set[str] = set()  # type: ignore
+    for plugin in allowed_plugins:
+        if plugin in seen:  # type: ignore
+            continue
+        yield plugin
+        seen.add(plugin)  # type: ignore
+        try:
+            yield from _get_allowed_plugins(dependency_lookup[plugin], dependency_lookup, seen)
```
jdkandersson, line 794: Generally try and limit the scope of `try...except` to the minimal code where the intended error could occur, e.g., this approach catches KeyErrors in the recursive call and on `dependency_lookup[plugin]`

src/jenkins.py
```diff
@@ -824,13 +825,22 @@ def _set_jenkins_system_message(message: str, container: ops.Container) -> None:
...
     """
     jcasc_yaml = container.pull(JCASC_CONFIG_FILE_PATH, encoding="utf-8").read()
     config = yaml.safe_load(jcasc_yaml)
-    config["jenkins"]["systemMessage"] = message
-    container.push(
-        JCASC_CONFIG_FILE_PATH, yaml.dump(config), encoding="utf-8", user=USER, group=GROUP
-    )
+    try:
+        config["jenkins"]["systemMessage"] = message
+        container.push(
```
jdkandersson, line 836: Should this be outside the `try...except`?

jdkandersson (approved):

src/status.py
```diff
@@ -7,6 +7,14 @@
 
 import ops
 
+PRIORITY_MAP = {
```
arturo-seijas, line 10: Would it make sense to open a ticket to request this in ops?

src/jenkins.py
```diff
@@ -768,69 +766,47 @@ def _build_dependencies_lookup(
...
-    seen.add(plugin)
-    for dependency in dependency_lookup[plugin]:
-        yield from _traverse_dependencies(dependency, dependency_lookup, seen)
-
-
 def _get_allowed_plugins(
-    top_level_plugins: typing.Iterable[str],
+    allowed_plugins: typing.Iterable[str],
     dependency_lookup: typing.Mapping[str, typing.Iterable[str]],
+    seen: set[str] | None = None,
```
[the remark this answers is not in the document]
yanksyoon: I think the dictionaries and sets in Python are references and initializing them in the argument would essentially make it a singleton which, in the context of the charm might be okay but can be a source of bugs.

src/status.py
yanksyoon: That sounds like a good idea :D I have seen that there is a [similar feature](https://ops.readthedocs.io/en/latest/#ops.CollectStatusEvent) but not exactly.

src/jenkins.py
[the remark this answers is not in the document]
arturo-seijas: Python ¬¬ TIL https://docs.python-guide.org/writing/gotchas/

arturo-seijas (approved):
LGTM!

src/jenkins.py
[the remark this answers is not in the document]
jdkandersson: yeh its a bad idea to use any mutable type as a default in an argument - unless you like investigating weird bugs

PR #26, opened 2023-08-22
chore: refactor update jenkins funcs

yanksyoon:
Applicable spec: N/A

### Overview

This PR 
- refactors functions related to Jenkins update into a single `update_jenkins` function.
- refactors injectable clients for testing to monkeypatched clients.
- refactors `remove_agent_node` function to take in the filesystem container rather than Jenkins credentials.

### Rationale

This PR is to address comments in previous PRs that may have been suited for a refactor PR to reduce complexity in a single PR and delivery time of other features.

### Juju Events Changes

None.

### Module Changes

Jenkins module has been refactored with an introduction of
- `has_lts_updates`: checks if there exists an update of the same LTS version that can be applied.
- `update_jenkins`: a wrapper function for functions related to updating Jenkins.
- `JenkinsRestartError`: to differentiate exceptions that is raised from `update_jenkins` function from blocking/non-blocking status.

### Library Changes

None.

### Checklist

- [x] The changes are compliant with [ISD054 - Manging Charm Complexity](https://discourse.charmhub.io/t/specification-isd014-managing-charm-complexity/11619)

The documentation does not require updating.

### DIFF
--- a/src/jenkins.py
+++ b/src/jenkins.py
@@ -738,7 +738,7 @@ def _get_plugin_name(plugin_info: str) -> str:
 
 def _build_desired_plugins_set(
     plugins_lookup: dict[str, list[str]], plugin: str, desired_plugins: set[str]
-):
+) -> None:
     """Build a set of plugins that are desired installations.
 
     Args:
@@ -753,7 +753,7 @@ def _build_desired_plugins_set(
         _build_desired_plugins_set(plugins_lookup, dependency, desired_plugins)
 
 
-def _set_jenkins_system_message(message: str, container: ops.Container):
+def _set_jenkins_system_message(message: str, container: ops.Container) -> None:
     """Set a system message on Jenkins.
 
     Args:
@@ -762,7 +762,6 @@ def _set_jenkins_system_message(message: str, container: ops.Container):
     """
     jcasc_yaml = container.pull(JCASC_CONFIG_FILE_PATH, encoding="utf-8").read()
     config = yaml.safe_load(jcasc_yaml)
-    logger.info("CONFIG: %s", config)
     config["jenkins"]["systemMessage"] = message
     container.push(
         JCASC_CONFIG_FILE_PATH, yaml.dump(config), encoding="utf-8", user=USER, group=GROUP
@@ -771,14 +770,14 @@ def _set_jenkins_system_message(message: str, container: ops.Container):
 
 def _build_dependencies_lookup(
     plugin_dependency_outputs: typing.Iterable[str],
-) -> dict[str, list[str]]:
-    """Build a lookup table of plugin short name to list of dependency plugin short name.
+) -> dict[str, tuple[str, ...]]:
+    """Build a lookup table of plugin short name to list of dependency plugin's short names.
 
     Args:
-        plugin_dependency_outputs (typing.Iterable[str]): _description_
+        plugin_dependency_outputs: The plugin dependency output from Jenkins Groovy script.
 
     Returns:
-        dict[str, list[str]]: _description_
+        The dependency lookup table.
     """
     dependency_lookup: dict[str, tuple[str, ...]] = {}
     for line in plugin_dependency_outputs:
@@ -797,7 +796,7 @@ def _build_dependencies_lookup(
 
 
 def _traverse_dependencies(
-    plugin: str, dependency_lookup: dict[str, typing.Iterable[str]], seen: set[str]
+    plugin: str, dependency_lookup: typing.Mapping[str, typing.Iterable[str]], seen: set[str]
 ) -> typing.Iterable[str]:
     """Traverse through plugin dependencies that have not yet been traversed recursively.
 
@@ -818,7 +817,8 @@ def _traverse_dependencies(
 
 
 def _get_allowed_plugins(
-    top_level_plugins: typing.Iterable[str], dependency_lookup: dict[str, typing.Iterable[str]]
+    top_level_plugins: typing.Iterable[str],
+    dependency_lookup: typing.Mapping[str, typing.Iterable[str]],
 ) -> typing.Iterable[str]:
     """Get the plugin short names of allowed plugins and its dependencies.
 
@@ -830,7 +830,7 @@ def _get_allowed_plugins(
     Yields:
         The allowed plugin short name.
     """
-    seen = set()
+    seen: set[str] = set()
     for plugin in top_level_plugins:
         if plugin in seen:
             continue
@@ -838,7 +838,7 @@ def _get_allowed_plugins(
 
 
 def _get_top_level_plugins(
-    plugins: typing.Iterable[str], dependency_lookup: dict[str, typing.Iterable[str]]
+    plugins: typing.Iterable[str], dependency_lookup: typing.Mapping[str, typing.Iterable[str]]
 ) -> typing.Iterable[str]:
     """Get top level plugins that are not dependencies to other plugins.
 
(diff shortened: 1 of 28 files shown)


src/jenkins.py
```diff
@@ -610,7 +597,34 @@ def get_updatable_version(proxy: state.ProxyConfig | None = None) -> str | None:
...
+
+    Raises:
+        JenkinsUpdateError: If there was an error fetching the Jenkins version information.
+
+    Returns:
+        True if an update within the same LTS is available. False otherwise.
+    """
+    try:
+        current_version = get_version()
+    except JenkinsError as exc:
```
arturo-seijas, line 614: I think you can merge this catch with the one on line 620 if you raise an exception[here](https://github.com/canonical/jenkins-k8s-operator/blob/d85ca11c8a48097bc409975ecf083c23c477d9df/src/jenkins.py#L555). This will potentially lead to a simpler error flow

src/jenkins.py
```diff
@@ -610,7 +597,34 @@ def get_updatable_version(proxy: state.ProxyConfig | None = None) -> str | None:
     return latest_version
 
 
-def download_stable_war(container: ops.Container, version: str) -> None:
+def has_lts_updates(proxy: state.ProxyConfig | None = None) -> bool:
```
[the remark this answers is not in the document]
jdkandersson: I think this function is intended to return whether the latest LTS version has a new patch version that can be installed, right? It does currently read like it is looking for whether there is a new LTS version

tests/unit/test_jenkins.py
```diff
@@ -322,38 +322,42 @@ def test__install_plugins(
...
 ):
     """
     arrange: given a test partial proxy config and a mock jenkins client.
     act: when _configure_proxy is called.
     assert: mock client is called with correct proxy arguments.
     """
-    mock_client = unittest.mock.MagicMock(spec=jenkinsapi.jenkins.Jenkins)
     mock_run_groovy_script = unittest.mock.MagicMock()
     mock_client.run_groovy_script = mock_run_groovy_script
+    monkeypatch.setattr(jenkins, "_get_client", lambda *_args, **_kwargs: mock_client)
```
jdkandersson, line 358: Shouldn't this use the client fixture?

jdkandersson: As in, can't the monkeypatch be done in the mock_client fixture?

src/jenkins.py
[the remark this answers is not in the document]
yanksyoon: Yes, I think it might make more sense within the context of the charm's `_update_jenkins_version` function, where the flow goes from `has_lts_updates` -> `update_jenkins` :), thank you for the comment!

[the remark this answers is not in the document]
jdkandersson: maybe `has_updates_for_lts`? That way it indicates that the check is whether there is an update for the lts version?

src/jenkins.py
```diff
@@ -635,6 +649,43 @@ def download_stable_war(container: ops.Container, version: str) -> None:
...
+    except JenkinsNetworkError as exc:
+        logger.error("Failed to download Jenkins executable, %s", exc)
+        raise JenkinsUpdateError("Error downloading Jenkins executable.") from exc
+
+    try:
+        safe_restart(container)
+        wait_ready()
+    except (JenkinsError, TimeoutError) as exc:
+        logger.error("Failed to restart Jenkins after updating, %s", exc)
+        raise JenkinsRestartError("Error restarting Jenkins after update.") from exc
```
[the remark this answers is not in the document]
yanksyoon: The reason is to differentiate between `BlockedStatus` from other statuses. JenkinsUpdateError could mean that the Jenkins is still alive while this function is being run but when Jenkins fails to restart, Jenkins is no longer providing service and hence should show `BlockedStatus`.

tests/unit/conftest.py
```diff
@@ -166,18 +162,18 @@ def register_command_handler(
...
     harness: Harness,
-    admin_credentials: Credentials,
+    admin_credentials: jenkins.Credentials,
     monkeypatch: pytest.MonkeyPatch,
     proxy_config: state.ProxyConfig,
 ) -> Container:
     """Harness Jenkins workload container that acts as a Jenkins container."""
     harness.set_can_connect("jenkins", True)
     container: Container = harness.model.unit.get_container("jenkins")
     container.push(
```
jdkandersson, line 172: I think there is a way to interact directly with the file system now in ops? So container push shouldn't be needed any more?

yanksyoon: I've pushed the change, it seems a lot more verbose.

src/jenkins.py
```diff
@@ -635,6 +649,43 @@ def download_stable_war(container: ops.Container, version: str) -> None:
...
+        JenkinsUpdateError: If there was an error updating Jenkins.
+        JenkinsRestartError: If there was an error restarting updated Jenkins.
+
+    Returns:
+        The updated Jenkins version.
+    """
+    try:
+        current_version = get_version()
+        latest_version = _get_latest_patch_version(current_version=current_version, proxy=proxy)
+    except (JenkinsNetworkError, JenkinsError, ValidationError) as exc:
```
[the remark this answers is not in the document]
yanksyoon: Merged and updated the error message, thank you for the feedback!

src/jenkins.py
yanksyoon: Merged and updated the error message, thank you for the review!

src/jenkins.py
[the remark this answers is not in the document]
yanksyoon: Updated the function name, thank you for the suggestion!

tests/unit/test_jenkins.py
yanksyoon: Saved a lot of lines of code, thanks for catching that!

jdkandersson (approved):

tests/unit/conftest.py
jdkandersson: The benefit is that how pathlib works is common knowledge so the code is easier to understand for the developer that is less familiar with container push etc

src/jenkins.py
```diff
@@ -611,15 +611,13 @@ def has_lts_updates(proxy: state.ProxyConfig | None = None) -> bool:
...
-    except JenkinsError as exc:
-        logger.error("Failed to get Jenkins version while fetching update, %s", exc)
-        raise JenkinsUpdateError("Failed to get Jenkins version.") from exc
-
-    try:
         latest_version = _get_latest_patch_version(current_version=current_version, proxy=proxy)
     except (JenkinsNetworkError, ValidationError) as exc:
         logger.error("Failed to fetch latest patch version info, %s", exc)
         raise JenkinsUpdateError("Failed to fetch latest patch version info.") from exc
+    except JenkinsError as exc:
```
arturo-seijas, line 618: I don't see any reason for having two except blocks here. It seems a little redundant considering the exceptions you are catching already have appropriate messages

src/jenkins.py
```diff
@@ -666,15 +664,10 @@ def update_jenkins(container: ops.Container, proxy: state.ProxyConfig | None = N
...
-    except (JenkinsNetworkError, JenkinsError, ValidationError) as exc:
-        logger.error("Failed to get Jenkins version metadata, %s", exc)
-        raise JenkinsUpdateError("Error fetching Jenkins version metadata.") from exc
-
-    try:
         _download_stable_war(container, latest_version)
-    except JenkinsNetworkError as exc:
-        logger.error("Failed to download Jenkins executable, %s", exc)
-        raise JenkinsUpdateError("Error downloading Jenkins executable.") from exc
+    except (JenkinsNetworkError, JenkinsError, ValidationError) as exc:
```
arturo-seijas, line 668: Isn't JenkinsNetworkError a subclass of JenkinsError?

src/jenkins.py
yanksyoon: Merged the error handlers, thank you!

merkata (approved):
good refactoring!

arturo-seijas (approved):
LGTM

PR #27, opened 2023-08-28
test: Git plugin integration test

yanksyoon:
Applicable spec: N/A

### Overview

This PR adds support for `git` plugin for Jenkins.

### Rationale

To test for commonly used plugins so that the list of plugins that are supported by `allowed-plugins` can have tested plugins.

### Juju Events Changes

None.

### Module Changes

Fixed a non-critical bug in `_build_dependencies_lookup` function which continues execution even if there were no dependencies found.

### Library Changes

None.

### Checklist

There is no documentation on Charmhub that needs updating at the moment. When we have more `default` allowed list of plugins, it can then be documented.


yanksyoon (commented):
Leaving comments to help with reviews

tests/integration/conftest.py
```diff
@@ -83,12 +83,12 @@ async def application_fixture(
         timeout=20 * 60,
         idle_period=30,
     )
-
     # slow down update-status so that it doesn't intervene currently running tests
-    async with ops_test.fast_forward(fast_interval="5h"):
-        yield application
-
-    await model.remove_application(application.name, force=True, block_until_done=True)
+    # `with ops_test.fast_forward` is not used here since juju cleanup will cause the tests to
```
jdkandersson, line 87: I'm not following this, it seems that the next line uses `with ops_test.fast_forward`?

tests/integration/conftest.py
```diff
@@ -365,35 +319,26 @@ def freeze_time_fixture() -> str:
...
 
-@pytest_asyncio.fixture(scope="function", name="restart_time_range_app")
-async def restart_time_range_app_fixture(application: Application):
+@pytest_asyncio.fixture(scope="function", name="prepare_restart_time_range")
+async def prepare_restart_time_range_fixture(application: Application):
     """Application with restart-time-range configured."""
     await application.set_config({"restart-time-range": "03-05"})
 
-    yield application
+    yield
```
jdkandersson, line 327: Should this maybe yield the value of the config that was set? It probably won't be used, just seems like a fixture should yield something

tests/integration/helpers.py
```diff
@@ -0,0 +1,179 @@
...
+        str(jenkins.PLUGINS_PATH),
+        "-p",
+        " ".join(plugins),
+    )
+    assert (
+        not returncode
+    ), f"Non-zero return code {returncode} received, stdout: {stdout} stderr: {stderr}"
+    # When there are connectivity issues, it prints in stderr and if not it prints in stdout.
+    assert any(
+        ("Done" in stdout, "Done" in stderr)
```
jdkandersson, line 54: If `Done` is in stderr, did the plugins actually install? Or do the connectivity issues cause the install to fail?

tests/integration/helpers.py
```diff
@@ -0,0 +1,179 @@
...
+    ), f"Failed to install plugins via juju ssh, {stdout}, {stderr}"
+
+    # the library will return 503 or other status codes that are not 200, hence restart and
+    # wait rather than check for status code.
+    jenkins_client.safe_restart()
+    model = ops_test.model
+    assert model, "Model not initialized."
```
jdkandersson, line 61: Should there be one fixture that does this? Not sure if there are other tests that need this check

tests/integration/test_k8s_agent.py
```diff
@@ -0,0 +1,80 @@
...
+
+async def test_jenkins_wizard_bypass(web_address: str):
+    """
+    arrange: given an active Jenkins charm's unit ip.
+    act: when web application is accessed
+    assert: wizard is bypassed and a login screen is shown.
+    """
+    response = requests.get(f"{web_address}/login", params={"from": "/"}, timeout=10)
+
+    assert "Unlock Jenkins" not in str(response.content)
```
jdkandersson, line 33: These can be a bit flaky since if the `Unlock Jenkins` message is changed but the login didn't work, this `not` will still pass. In these cases it can help to put a check at the beginning of the test which checks that the message does appear and then follow that up with the action that should make it disappear and check it is gone

tests/integration/helpers.py
yanksyoon: Sometimes the "Done" is printed on the stderr if there have been connectivity issues and it tries some other mirrors/links. As long as "Done" is in the string, it is installed correctly.

jdkandersson: perhaps clarify that in the comment? The current comment doesn't seem to include the details that it tries other mirrors/ links

tests/integration/conftest.py
yanksyoon: This is the current line under test, this will either be finalized with `with ops_test.fast_forward` of `model.set_config`, whichever one fails the tests less.

tests/integration/test_k8s_agent.py
yanksyoon: This does not test login but the setup of Jenkins which bypasses the wizard. The wizard contains the string "Unlock Jenkins" whereas the unlocked Jenkins contains the string "Welcome to Jenkins". Jenkins charm, when in `maintenance` status while setting up, can display "Unlock Jenkins" wizard page which should be resolved when the charm does become `Active`. 
The reason for testing for "Unlock Jenkins" being in the response content is to mark the fact that this message will appear when the setup wizard has not been bypassed. 
Since this isn't quite clear in the testing code, I've added comment to resolve this issue.
Thank you for the comment!

tests/integration/helpers.py
yanksyoon: Added comment, thank you!

tests/integration/conftest.py
yanksyoon: Yielded application and renamed this to `app_with_restart_config` as a fixture, thank you!

tests/integration/test_plugins.py
```diff
@@ -34,10 +39,37 @@ async def test_jenkins_plugins_config(
...
+    # check that git plugin git repository validation works on Jenkins server
+    check_url_res = unit_web_client.client.requester.post_url(
+        f"{unit_web_client.client.baseurl}/job/{job_name}/descriptorByName/"
+        "hudson.plugins.git.UserRemoteConfig/checkUrl",
+        data={
+            "value": "https://github.com/canonical/jenkins-k8s-operator",
+            "credentialsId": "",
+        },
+    )
+    assert (
```
amandahla, line 73: I'm not very familiar with Jenkins, so just to get the idea: You submit the job using the post_url with the data and then you know that it was completed if there is a "<div/>" in the content response, is that right?

yanksyoon: Yup, it returns any error messages in HTML, in this case, there were no errors and hence an empty `<div />` is returned.

tests/integration/conftest.py
yanksyoon: I've decided to stick with `op_test.fast_forward`, the comments have been updated accordingly.
```
# slow down update-status so that it doesn't intervene currently running tests
    # don't yield inside the context since juju cleanup is not reliable.
    # model.set_config(...) also doesn't work as well as the following code.
    async with ops_test.fast_forward(fast_interval="5h", slow_interval="5h"):
        pass
    yield application
```

yhaliaw (approved):
LGTM

jdkandersson (approved):

PR #36, opened 2023-09-22
chore: trivy ignores (CI)

yanksyoon:
Applicable spec: N/A

### Overview

Add Jenkins XSS vulnerability and permissions vulnurability to trivyignores file.

### Rationale

To get CI actions to pass.

### Juju Events Changes

None.

### Module Changes

None.

### Library Changes

None.

### Checklist


.trivyignore
```diff
@@ -10,6 +10,8 @@ CVE-2023-35141
 CVE-2023-2976
 CVE-2023-39151
 CVE-2023-34034
+CVE-2023-43495
+CVE-2023-43496
```
jdkandersson, line 14: Do you think it would be worth adding documentation here why these should be ignored rather than us doing something to resolve/ mitigate the CVE? If there is a CVE we should always consider whether we can do something about it and it would also be good to know when the CVE might be resolved upstream so that when we check back here we know whether we can remove the CVE

yanksyoon: One thing that I've got in my notes is to update the Jenkins version pre-release and remove (hopefully) most of the CVEs here. 
Would creating a backlog for it be alright? The change would need to:
1. Remove automatic patch updates
2. Update Jenkins version
3. Remove resolved CVEs

jdkandersson: Yes if it includes documenting the existing CVEs for why we aren't doing anything about them 🙂

jdkandersson (approved):

weiiwang01 (approved):

PR #51, opened 2023-11-06
feat: rotate-credentials action

yanksyoon:
Applicable spec: N/A

### Overview

Adds a rotate credential action, which invalidates all user sessions and re-generates the admin user password.

### Rationale

For data breach cases.

### Juju Events Changes

- Added `rotate_credentials_action` event listener

### Module Changes

- Refactored `get_admin_password_action` event listener into a separate actions observer.

### Library Changes

None.

### Checklist


src/jenkins.py
```diff
@@ -3,11 +3,14 @@
 
 """Functions to operate Jenkins."""
 
+# pylint:disable=too-many-lines
+
```
yanksyoon, line 7: Refactoring of the jenkins module should be done in another PR, currently out of scope.

docs/how-to/rotate-credentials.md
```diff
@@ -0,0 +1,28 @@
+# How to rotate credentials
+
+### Rotate credentials
+
+To rotate credentials for the admin account and invalidate all logged in user sessions, you can run
```
arturo-seijas, line 5: s/you can run/run

docs/how-to/rotate-credentials.md
```diff
@@ -0,0 +1,28 @@
...
+the `rotate-credentials` action.
+
+```
+juju run-action jenkins-k8s/0 rotate-credentials --wait
+```
+
+The output should look something similar to the contents below:
```
arturo-seijas, line 12: s/look something similar/look similar

tests/integration/test_jenkins.py
```diff
@@ -98,3 +99,23 @@ async def test_jenkins_automatic_update(
...
     assert updated_workload_version == latest_jenkins_lts_version, "The Jenkins should be updated."
+
+
+async def test_rotate_password_action(unit_web_client: UnitWebClient):
+    """
+    arrange: given a jenkins API session that is connected.
+    act: when rotate password action is called.
+    assert: the session is invalidated and new password is returned.
+    """
+    sess = unit_web_client.client.requester.session
```
arturo-seijas, line 110: s/sess/session

tests/integration/test_jenkins.py
```diff
@@ -98,3 +99,23 @@ async def test_jenkins_automatic_update(
...
+    sess.get(f"{unit_web_client.web}/manage/")
+    action: Action = await unit_web_client.unit.run_action("rotate-credentials")
+    await action.wait()
+    new_password: str = action.results["password"]
+
+    assert unit_web_client.client.password != new_password, "Password not rotated"
+    res = sess.get(f"{unit_web_client.web}/manage/")
```
arturo-seijas, line 117: s/res/result

tests/unit/test_jenkins.py
```diff
@@ -1506,3 +1506,65 @@ def test_remove_unlisted_plugins(  # pylint: disable=too-many-arguments
         )
     else:
         mock_client.delete_plugins.assert_not_called()
+
+
+def test__invalidate_sessions(
```
arturo-seijas, line 1511: Is it me or there are two consecutive underscores right after `test` in these method names?

arturo-seijas, line 1511: I can't see the point of adding this test

tests/unit/test_jenkins.py
```diff
@@ -1506,3 +1506,65 @@ def test_remove_unlisted_plugins(  # pylint: disable=too-many-arguments
...
+    act: when _invalidate_sessions is called.
+    assert: no errors are raised.
+    """
+    mock_client.run_groovy_script = lambda *_args, **_kwargs: None
+    monkeypatch.setattr(jenkins, "_get_client", lambda *_args, **_kwargs: mock_client)
+
+    jenkins._invalidate_sessions(container)
+
+
+def test__set_new_password(
```
arturo-seijas, line 1525: Same here

tests/unit/test_jenkins.py
yanksyoon: This was solely done to increase the coverage tbh.

tests/unit/test_actions.py
```diff
@@ -0,0 +1,130 @@
...
+import pytest
+
+import charm
+import jenkins
+from charm import JenkinsK8sOperatorCharm
+
+from .types_ import HarnessWithContainer
+
+
+def test__on_get_admin_password_action_container_not_ready(
```
arturo-seijas, line 23: Really picky, but how about `test_get_admin_password_action_container_not_ready` to avoid referencing a private method?

tests/unit/test_jenkins.py
yanksyoon: This was to increase coverage as well, i could opt in for just skipping these lines.

tests/unit/test_actions.py
yanksyoon: Done, thanks for the suggestion!

arturo-seijas: I think you forgot to remove the double underscores

yanksyoon: Woops, done! Thanks!

arturo-seijas (approved):
LGTM

src/actions.py
```diff
@@ -0,0 +1,60 @@
...
+
+"""Jenkins charm actions."""
+
+import ops
+
+import jenkins
+from state import State
+
+
+class Observer(ops.Object):
```
amandahla, line 12: That looks a little bit different from what is recommended by the Charm Complexity guide.
Flask, for example, observe action event in the charm logic
https://github.com/canonical/flask-k8s-operator/blob/15056fd2451a5279ab45880dee5fa6fca315bae0/src/charm.py#L94

actions.yaml
```diff
@@ -4,3 +4,6 @@
 get-admin-password:
   description: |
     Retrieve the admin password which has been auto-generated during Jenkins install.
+rotate-credentials:
+  description: |
+    Invalidate all current sessions and generate a new password for admin account.
```
amandahla, line 9: Out of curiosity: why there is one action to do two things instead of two separate actions like invalidate-all-sessions and generate-admin-password ?

yanksyoon: This is taken from an action in wordpress as well, where this action is used to re-secure the application from a data breach - invalidate all the logged in users and rotate the password. So for example, when the admin user is breached, unless we log out that user, the admin user will still be able to cause damage while the session lasts.

src/actions.py
yanksyoon: I'm not sure I can find actions under [Managing Charm Complexity](https://discourse.charmhub.io/t/specification-isd014-managing-charm-complexity/11619) document, perhaps have I missed something?

As of my reasoning, I tried to make things as modular as possible so that it's smaller in scope. One of the problems I'm facing now is a huge `jenkins.py` file which i had to slap on a `# pylint: ignore`. Would that be a good enough reasoning to make this into a separate module? I dont' mind moving actions back into the main `charm.py` :)

Thank you for the review, +1 review :coin:

actions.yaml
amandahla: I'm not familiar with Jenkins so I was wondering if this is usually needed this way instead of creating two actions :-)

src/actions.py
amandahla: I would go for leaving in the charm. What do you think @jdkandersson ?

tests/integration/test_jenkins.py
```diff
@@ -98,3 +99,23 @@ async def test_jenkins_automatic_update(
...
+    action: Action = await unit_web_client.unit.run_action("rotate-credentials")
+    await action.wait()
+    new_password: str = action.results["password"]
+
+    assert unit_web_client.client.password != new_password, "Password not rotated"
+    result = session.get(f"{unit_web_client.web}/manage/")
+    assert result.status_code == 403, "Session not cleared"
+    new_client = jenkinsapi.jenkins.Jenkins(unit_web_client.web, "admin", new_password)
+    result = new_client.requester.get_url(f"{unit_web_client.web}/manage/")
+    assert result.status_code == 200, "Invalid password"
```
yhaliaw, line 121: Should whether the new password is working be tested?

yanksyoon: The line `new_client = jenkinsapi.jenkins.Jenkins(unit_web_client.web, "admin", new_password)` will fail to instantiate when the new password doesn't work :)

yhaliaw (approved):
LGTM.
I feel the actions could be in it own file, since the charm.py is already fairly large.
I am fine either way.

arturo-seijas (dismissed):

src/actions.py
jdkandersson: It seems reasonable to have an actions observer to reduce the amount of code in `charm.py`. We don't need to do it everywhere, e.g., if `charm.py` is quite short (like 100-200 lines), the action observation can be done there. If it is starting to get complex, this can be one of the ways that code can be moved to a separate module

src/jenkins.py
jdkandersson: This seems to be something that is up for debate - usually I try to fix these kind of linting issues the first time they come up. If I find myself spending too much time on it or it increases the complexity of the PR by a lot then it can be left to a future PR, otherwise it can be reasonable to do it in the same PR.

The solution here could be just to split this file into two or three separate files with logically grouped functions.

Systematically leaving these kind of issues in a PR can lead to a lot of follow up PRs that we may forget to get to. Most other linting issues introduced in a PR are usually fixed within a PR

jdkandersson: Another way to address this is to do it in a follow up PR and associate it with the same Jira ticket/ issue so that we don't call the feature done until the linting issues introduced by it have been resolved

yanksyoon: @jdkandersson This can now be removed since the later PR #60 has already been merged and has reduced the LOC significantly in `jenkins.py`. :) Updating now!

tests/integration/test_jenkins.py
```diff
@@ -61,6 +61,26 @@ async def test_jenkins_automatic_update_out_of_range(
...
 
 
+async def test_rotate_password_action(unit_web_client: UnitWebClient):
+    """
+    arrange: given a jenkins API session that is connected.
+    act: when rotate password action is called.
+    assert: the session is invalidated and new password is returned.
+    """
+    session = unit_web_client.client.requester.session
+    session.get(f"{unit_web_client.web}/manage/")
```
jdkandersson, line 71: Should we check that a 200 is returned here?

tests/unit/test_actions.py
```diff
@@ -0,0 +1,129 @@
...
+):
+    """
+    arrange: given a jenkins container that is not connectable.
+    act: when get-admin-password action is run.
+    assert: the event is failed.
+    """
+    harness_container.harness.set_can_connect(
+        harness_container.harness.model.unit.containers["jenkins"], False
+    )
+    mock_event = MagicMock(spec=ops.ActionEvent)
```
jdkandersson, line 33: @benhoyt it could be interesting to make this easier, like providing a fake event class or something like that for use in tests?

jdkandersson: It is relatively minor, it would be handy for the framework to set the `spec` argument rather than it being done by users of the framework so that we are sure that we're using the right `spec` class

tests/unit/test_actions.py
```diff
@@ -0,0 +1,129 @@
...
+    harness_container.harness.begin()
+
+    jenkins_charm = typing.cast(JenkinsK8sOperatorCharm, harness_container.harness.charm)
+    jenkins_charm.actions_observer.on_get_admin_password(mock_event)
+
+    mock_event.set_results.assert_called_once_with({"password": admin_credentials.password})
+
+
+def test_on_rotate_credentials_action_container_not_ready(
```
jdkandersson, line 59: @benhoyt also something to consider, it seems that not being able to connect to the container would probably fail many of the action events, something to consider whether the framework could provide guarantees, a decorator or something that makes this less intensive to code and then to test within charms for each of the actions

tests/unit/test_jenkins.py
```diff
@@ -1108,3 +1108,37 @@ def test_remove_unlisted_plugins(  # pylint: disable=too-many-arguments
...
+    arrange: given a monkeypatched _invalidate_sessions that raises JenkinsAPIException.
+    act: when rotate_credentials is called.
+    assert: JenkinsError is raised.
+    """
+    monkeypatch.setattr(
+        jenkins,
+        "_invalidate_sessions",
+        lambda *_args, **_kwargs: raise_exception(
+            jenkinsapi.custom_exceptions.JenkinsAPIException
+        ),
```
jdkandersson, line 1126: Perhaps instead of a `lambda` just use a mock with a side effect of raising an exception?

jdkandersson: Also applies to the rest of the `setattr`s with `lmabda`s

jdkandersson:
Btw, this is quite a good PR, well coded!

tests/unit/test_jenkins.py
yanksyoon: Pushed on a separate PR to ease the reviews https://github.com/canonical/jenkins-k8s-operator/pull/62, thank you for the review!

merkata (dismissed):

arturo-seijas (approved):

tests/unit/test_actions.py
benhoyt: If I understand the test correctly we're just testing an action here (and that it called `fail`). The new support for testing actions with `Harness.run_action` should be good for this. See https://github.com/canonical/operator/pull/1053

@tonyandrewmeyer added this recently. It's merged in the `main` branch, and will be included in the next `ops` release at the end of November, 2.9.0.

tests/unit/test_actions.py
benhoyt: It seems to me the test is doing too much here. There's no need to call `set_can_connect` with `False` as that's the default for containers under `Harness` anyway (when using `begin` and not `begin_with_initial_hooks`).

Additionally, I just looked at the code for this action:

```python
    def _on_get_admin_password(self, event: ops.ActionEvent) -> None:
        container = self.unit.get_container(self.state.jenkins_service_name)
        if not container.can_connect():
            event.defer()
            return
        credentials = jenkins.get_admin_credentials(container)
        event.set_results({"password": credentials.password_or_token})
```

It doesn't seem right at all to use `defer()` in an `ActionEvent`. In fact, `ActionEvent.defer` always raises a `RuntimeError` -- so I'm not sure how this code path is working at all (maybe it's not and the test isn't actually testing what we think it is). We definitely shouldn't use `defer` inside an action; this should be changed to an `event.fail()` or similar (or just let it error?).

tests/unit/test_jenkins.py
yanksyoon: Merged to this branch.

tests/unit/test_actions.py
yanksyoon: The test has been simplified with harness (`harness_with_container` fixture by default enables jenkins container to be ready to reduce repeating set_can_connect).

For the later code snippet, the event defer has been removed a few PRs back :)
Thank you for the review!

tests/unit/conftest.py
```diff
@@ -253,22 +253,24 @@ def harness_container_fixture(harness: Harness, container: Container) -> Harness
     return HarnessWithContainer(harness=harness, container=container)
 
 
-@pytest.fixture(scope="function", name="raise_exception")
-def raise_exception_fixture():
+@pytest.fixture(scope="function", name="raise_exception_mock")
+def raise_exception_mock_fixture():
```
jdkandersson, line 257: It is quite easy to create a mock object that raises an exception when it is called:

```
>>> from unittest import mock
>>> test = mock.MagicMock(side_effect=Exception)
>>> test()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.10/unittest/mock.py", line 1114, in __call__
    return self._mock_call(*args, **kwargs)
  File "/usr/lib/python3.10/unittest/mock.py", line 1118, in _mock_call
    return self._execute_mock_call(*args, **kwargs)
  File "/usr/lib/python3.10/unittest/mock.py", line 1173, in _execute_mock_call
    raise effect
Exception
```

Is it worth creating a fixture given the mock creation is a single line of code?

yanksyoon: Fixed, thank you!
Also fixed consistency in imports

jdkandersson (approved):

src/charm.py
```diff
@@ -50,14 +51,11 @@ def __init__(self, *args: typing.Any):
         except CharmRelationDataInvalidError as exc:
             raise RuntimeError("Invalid relation data received.") from exc
 
+        self.actions_observer = actions.Observer(self, self.state)
         self.agent_observer = agent.Observer(self, self.state)
         self.cos_observer = cos.Observer(self)
         self.ingress_observer = ingress.Observer(self)
         self.framework.observe(self.on.jenkins_pebble_ready, self._on_jenkins_pebble_ready)
-        self.framework.observe(self.on.get_admin_password_action, self._on_get_admin_password)
-        self.framework.observe(
```
Thanhphan1147, line 58: Coming from [this PR](https://github.com/canonical/jenkins-k8s-operator/pull/68). I think this line is necessary so that the `_on_storage_attached` hook is run to set the proper ownership when storage is attached

Thanhphan1147 (changes requested):

src/charm.py
yanksyoon: Just realized it got deleted during the merge conflict fixes, added back! Thanks!

src/actions.py
amandahla: Considering the last answer, I understood that in this case this should be moved to charm.py, right?

jdkandersson: I'll leave that to you and Yanks to decide 🙂

src/charm.py
jdkandersson: @Thanhphan1147 please see https://github.com/canonical/is-charms-contributing-guide/blob/main/CONTRIBUTING.md#pr-comments-and-requests-for-changes for guidance on requesting changes in PRs

jdkandersson (approved):

amandahla (commented):
LGTM

amandahla (approved):
LGTM

src/actions.py
amandahla: No worries then :)

Thanhphan1147 (approved):
LGTM, thanks

PR #68, opened 2023-11-22
[WIP] feat: precondition check (fix/storage mount)

yanksyoon:
Applicable spec: N/A

### Overview

Adds a charm precondition checker: storage mounted, container ready before taking any service related actions.

### Rationale

- Removes repeated code checking for container ready.
- Allows safe attachment of storage before the service is initialized.

### Juju Events Changes

None.

### Module Changes

Added `precondition.py`

### Library Changes

None.

### Checklist


src/charm.py
```diff
@@ -51,6 +52,12 @@ def __init__(self, *args: typing.Any):
...
 
+        try:
+            precondition.check(self, self.state)
+        except precondition.ConditionCheckError as exc:
+            self.unit.status = ops.WaitingStatus(exc.msg)
+            return
+
         self.actions_observer = actions.Observer(self, self.state)
         self.agent_observer = agent.Observer(self, self.state)
         self.cos_observer = cos.Observer(self)
```
Thanhphan1147, line 63: Can you confirm that this [line](https://github.com/canonical/jenkins-k8s-operator/blob/3e4c253492ea1f8fa7c0a33767cbaff515cc5905/src/charm.py#L58) is missing in `charm.py`? I find it strange that it's not present in the diff :confused:

Thanhphan1147 (approved):

src/charm.py
Thanhphan1147: Without it the hook won't run so I'm wondering why the tests are passing

src/charm.py
```diff
@@ -96,17 +103,9 @@ def _get_pebble_layer(self, jenkins_env: jenkins.Environment) -> ops.pebble.Laye
...
-            event: The event fired when pebble is ready.
-        """
-        container = event.workload
-        if not container or not container.can_connect():
-            event.defer()
-            return
-
+    def _on_jenkins_pebble_ready(self, _: ops.PebbleReadyEvent) -> None:
+        """Configure and start Jenkins server."""
+        container = self.unit.get_container(self.state.jenkins_service_name)
```
Thanhphan1147, line 108: I'm wondering how this will affect events that are only fired once. Do we know if `pebble_ready` will be fired again if we don't do `event.defer()` ? I have the same question for `_on_storage_attached`

yanksyoon: Right, that is true, it seems like the precondition check may need to be called for every hook events as well. Let me write a spec for that as @jdkandersson has suggested. Turning this PR into draft for now. Thanks for the review!

src/charm.py
yanksyoon: It is probably due to the merge going into `feat/rotate-password` rather than `main`, I branched off from `feat/rotate-password` because there are quite a few changes that will lead to merge conflicts. Let me merge the `feat/rotate-password` with main and then merge this with updated `feat/rotate-password` :D

yanksyoon: Nvm, it was removed during merge conflict fixes, adding them back! sorry!

yanksyoon:
Quickfix without implementing precondition pattern has been merged. Closing.

PR #69, opened 2023-11-22
Changed hard-coded value to constant

Thanhphan1147:
minor improvements to an integration test

instead of `assert "2000" in str(action.results.get("stdout"))` we define a constant in `./src/jenkins` and use that for code readability
Check for both user and group ownership instead of just user

### Checklist


src/jenkins.py
```diff
@@ -54,6 +54,8 @@
 
 USER = "jenkins"
 GROUP = "jenkins"
+UID = "2000"
```
arturo-seijas, line 57: These values are only used in tests. Can you define them in the test sources instead?

yanksyoon (approved):

src/jenkins.py
Thanhphan1147: Thanks, I moved them to the test file

yanksyoon (approved):

amandahla (approved):
LGTM

PR #70, opened 2023-11-23
fix: juju 3.1 compatible cos stack

yanksyoon:
Applicable spec: N/A

### Overview

To fix integration tests that rely on juju 2.9, the latest COS is not backwards compatible with juju 2.9.

### Rationale

Our operator workflows use juju 2.9 and many of the prod environments are in 2.9.

### Juju Events Changes

None.

### Module Changes

None.

### Library Changes

None.

### Checklist

### DIFF
--- a/.github/workflows/integration_test.yaml
+++ b/.github/workflows/integration_test.yaml
@@ -8,7 +8,7 @@ jobs:
     uses: canonical/operator-workflows/.github/workflows/integration_test.yaml@main
     secrets: inherit
     with:
-      channel: 1.26-strict/stable
+      channel: 1.28-strict/stable
       extra-arguments: |
         --kube-config ${GITHUB_WORKSPACE}/kube-config
       modules: '["test_jenkins.py", "test_k8s_agent.py", "test_machine_agent.py", "test_plugins.py", "test_proxy.py", "test_cos.py"]'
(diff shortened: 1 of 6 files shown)


mthaddon:
This charm isn't deployed anywhere in production yet, is it? If not, I think it'd be better for us to focus on only supporting juju 3.1 and later. IS have plans to deploy Juju 3 controllers in PS5 this cycle, and any new deployments I think can rely on being on Juju 3. I realise that may mean some extra work now to get our workflows working with Juju 3, but I'm not sure we should invest time in making COS work on Juju 2.9 because (for instance) even if you pin the charm to a specific revision, the OCI image resource will always be the latest in that channel, so it might well not work as expected.

arturo-seijas (approved):

arturo-seijas (approved):

.github/workflows/integration_test.yaml
```diff
@@ -8,11 +8,12 @@ jobs:
     uses: canonical/operator-workflows/.github/workflows/integration_test.yaml@main
     secrets: inherit
     with:
+      channel: 1.26-strict/stable
```
jdkandersson, line 11: I'm guessing this is microk8s, why not use 1.28-strict/stable?

yanksyoon: Updated! Thank you!

jdkandersson (approved):

yhaliaw (approved):
LGTM

PR #75, opened 2023-12-05
fix: prevent hooks before preconditions ready

yanksyoon:
Applicable spec: N/A

### Overview

Prevent any service events (pebble_ready, storage_ready, update_status, actions) from continuing before all preconditions (container ready & storage mounted) are ready.

### Rationale

Prevent unstable behavior of Jenkins when the service starts before the storage has been properly mounted.

### Juju Events Changes

N/A

### Module Changes

`charm.py` - event handlers have storage attached check.
`actions.py` - action event handlers have storage attached check.
`agent.py` - agent relation handlers have storage attached check.

### Library Changes

None.

### Checklist

### DIFF
--- a/src/actions.py
+++ b/src/actions.py
@@ -36,8 +36,8 @@ class Observer(ops.Object):
             event: The event fired from get-admin-password action.
         """
         container = self.charm.unit.get_container(self.state.jenkins_service_name)
-        if not container.can_connect():
-            event.fail("Container not yet ready.")
+        if not container.can_connect() or not self.model.storages.get(self.state.storage_name):
+            event.fail("Service not yet ready.")
             return
         credentials = jenkins.get_admin_credentials(container)
         event.set_results({"password": credentials.password_or_token})
@@ -49,8 +49,8 @@ class Observer(ops.Object):
             event: The rotate credentials event.
         """
         container = self.charm.unit.get_container(self.state.jenkins_service_name)
-        if not container.can_connect():
-            event.fail("Container not yet ready.")
+        if not container.can_connect() or not self.model.storages.get(self.state.storage_name):
+            event.fail("Service not yet ready.")
             return
         try:
             password = jenkins.rotate_credentials(container)
--- a/src/actions.py
+++ b/src/actions.py
@@ -3,21 +3,16 @@
 
 """Jenkins charm actions."""
 
-import typing
-
 import ops
 
 import jenkins
-from state import State
-
-if typing.TYPE_CHECKING:
-    from charm import JenkinsK8sOperatorCharm  # pragma: no cover
+from state import JENKINS_SERVICE_NAME, State
 
 
 class Observer(ops.Object):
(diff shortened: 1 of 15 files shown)


src/actions.py
```diff
@@ -36,8 +36,8 @@ def on_get_admin_password(self, event: ops.ActionEvent) -> None:
             event: The event fired from get-admin-password action.
         """
         container = self.charm.unit.get_container(self.state.jenkins_service_name)
-        if not container.can_connect():
-            event.fail("Container not yet ready.")
+        if not container.can_connect() or not self.model.storages.get(self.state.storage_name):
```
jdkandersson, line 39: Let's check whether this guarantees access, let's use the wordpress approach for now

yanksyoon: Added /proc/mount check, thank you!

src/charm.py
```diff
@@ -61,6 +65,19 @@ def __init__(self, *args: typing.Any):
...
+        """Return whether the Jenkins home storage is mounted.
+
+        Returns:
+            True if storage is mounted, False otherwise.
+        """
+        container = self.unit.get_container(self.state.jenkins_service_name)
+        if not container.can_connect():
+            return False
+        mount_info: str = container.pull("/proc/mounts").read()
+        return str(jenkins.HOME_PATH) in mount_info
```
jdkandersson, line 79: Can't we make this a shared function that takes the container as an argument so we don't need to pass the charm around to other classes?

yanksyoon: @jdkandersson  Out of curiosity, would there be a reason why passing charm around would not be preferred? I was thinking that `charm.py` could provide utility functions that could be shared around so that such functions don't need to be repeated and be readily available around the observers.

jdkandersson: Because of circular dependencies which is why the charm is no longer typed as a base charm and instead as the specific charm, so it depends on its creator which is difficult to reason about

yanksyoon: The circular dependency is due to the typing - no runtime dependency - I'm not quite sure I understand why coupling here would be difficult to reason about.

yanksyoon: Updated, thank you!

src/state.py
```diff
@@ -214,6 +217,22 @@ def from_env(cls) -> typing.Optional["ProxyConfig"]:
...
+        charm: The Jenkins k8s charm.
+
+    Returns:
+        True if storage is mounted, False otherwise.
+    """
+    container = charm.unit.get_container(JENKINS_SERVICE_NAME)
+    if not container.can_connect():
+        return False
+    mount_info: str = container.pull("/proc/mounts").read()
+    return str(jenkins.HOME_PATH) in mount_info
```
jdkandersson, line 233: Do you think this is a private function? I'm guessing it should only be used to construct the charm state?

yanksyoon: Fixed! thank you :)

jdkandersson (approved):

cbartz (approved):
LGTM

PR #82, opened 2023-12-08
fix: plugin remove

yanksyoon:
Applicable spec: N/A

### Overview

When a user is installing plugins through the UI and by chance the update-status hook runs (which does plugins auto-remove), there is a race condition.
While the UI installs plugins, the remove-plugins generates an incomplete view of all the plugins and removes all the plugins that are to be installed.

### Rationale

To wait for all plugins downloads to be settled before auto-removing plugins.

### Juju Events Changes

None.

### Module Changes

`jenkins.py`'s `remove_unlisted_plugins` now waits for plugins to complete donwload by checking for existence of .tmp files.

### Library Changes

None.

### Checklist


tests/unit/test_jenkins.py
jdkandersson, line 1: Would it also be interesting to have an integration test which simulates a plugin being installed and then triggering the check for whether plugins should be removed?

cbartz (commented):
Looks good, but as mentioned in jdkandersson's comment, an integration test might be a good addition.

src/charm.py
```diff
@@ -163,8 +163,8 @@ def _remove_unlisted_plugins(self, container: ops.Container) -> ops.StatusBase:
             logger.error("Failed to remove unlisted plugin, %s", exc)
             return ops.StatusBase.from_name(original_status, "Failed to remove unlisted plugin.")
         except TimeoutError as exc:
-            logger.error("Failed to restart jenkins after removing plugin, %s", exc)
-            return ops.BlockedStatus("Failed to restart Jenkins after removing plugins")
+            logger.error("Failed to remove plugins, %s", exc)
+            return ops.BlockedStatus("Failed to remove plugins.")
```
amandahla, line 167: Since is a blocked status, what would be the action that the user can take to solve this?

src/jenkins.py
```diff
@@ -607,6 +607,23 @@ def _get_plugin_name(plugin_info: str) -> str:
...
+
+    This function checks for any .tmp files in the plugins directory which indicates that a user
+    might be installing plugins through the UI.
+
+    Args:
+        container: The Jenkins workload container.
+        timeout: Timeout in seconds to wait for plugins to be installed.
+    """
+    _wait_for(
+        lambda: not container.list_files(path="/var/lib/jenkins/plugins/", pattern="*.tmp"),
```
amandahla, line 621: What happens if list_files fails?

tests/unit/test_jenkins.py
yanksyoon: Added an integration test that:
1. Starts plugins installations via UI
2. Waits for plugin downloads to begin (.tmp files in plugins dir)
3. Fires update-status event
4. Checks for warning log ("Plugins being downloaded, waiting until further actions.") in `juju debug-log`
5. Checks that the plugins are correctly installed after restart

src/jenkins.py
yanksyoon: It should raise a PathError which should make the charm fall into an ErroredStatus. The charm should not be able to trigger this before the installation of Jenkins so if this path is missing, it should fall into an ErroredStatus :D

src/charm.py
yanksyoon: I think there was a discussion around `DegradedStatus` but it isn't quite supported yet. This is to indicate that the service may be down or degraded.
In order to resolve it, the user must either:
1. Restart Jenkins via pebble again if Jenkins has failed to restart for some reason
2. Resolve the plugin remove failure (which is usually due to dependency problems)

jdkandersson (approved):

src/charm.py
amandahla: Nice, do you think it could be described in the message or log?

src/jenkins.py
amandahla: I'm not familiar if this situation might easily happens or not but do you think is worthy creating a unit test for that case?

src/charm.py
yanksyoon: WIll create a followup PR since it's break right now :D

src/jenkins.py
yanksyoon: WIll create a followup PR since it's break right now, the integration test has been added for a bigger scope for now :)

PR #85, opened 2023-12-12
fix: integration test & workflow

yanksyoon:
Applicable spec: N/A

### Overview

!This is a draft PR!

1. The workflow fails due to insufficient permissions.
2. The integration tests fails due to an update with blueocean plugin and a bug in jenkins-plugins-manager which doesn't resolve the dependency correctly and one of the dependent plugin downstream is not downloaded which causes the blueocean plugin to break.

### Rationale

To make the integration tests pass.

### Juju Events Changes

none.

### Module Changes

none.

### Library Changes

none.

### Checklist

- [ x] The [charm style guide](https://juju.is/docs/sdk/styleguide) was applied

### DIFF
--- a/.github/workflows/integration_test.yaml
+++ b/.github/workflows/integration_test.yaml
@@ -21,4 +21,4 @@ jobs:
         ./tests/integration/pre_run_script.sh"
       juju-channel: 3.1/stable
       self-hosted-runner: true
-      self-hosted-runner-label: "edge"
+      self-hosted-runner-label: "xlarge"
--- a/.github/workflows/integration_test.yaml
+++ b/.github/workflows/integration_test.yaml
@@ -7,6 +7,9 @@ jobs:
   integration-tests:
     uses: canonical/operator-workflows/.github/workflows/integration_test.yaml@main
     secrets: inherit
+    permissions:
+      contents: read
+      packages: write
     with:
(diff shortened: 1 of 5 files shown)


.github/workflows/integration_test.yaml
```diff
@@ -7,6 +7,9 @@ jobs:
   integration-tests:
     uses: canonical/operator-workflows/.github/workflows/integration_test.yaml@main
     secrets: inherit
+    permissions:
+      contents: read
+      packages: write
```
jdkandersson, line 12: Is this still needed? I think they fixed the settings

yanksyoon: https://github.com/canonical/operator-workflows/blob/main/.github/workflows/build_rocks.yaml#L82

I think it is required due to this line, otherwise it would throw an invalid workflow error (example to the run below). 
https://github.com/canonical/jenkins-k8s-operator/actions/runs/7169188246

I'm not sure if this is also fixed by enterprise level permissions, i'll remove and try running it as well.

jdkandersson: Yeh I saw somewhere that it should be fixed now

yhaliaw (approved):

jdkandersson (approved):

PR #86, opened 2023-12-14
fix: agent connection through direct tunnel

yanksyoon:
Applicable spec: N/A

### Overview

When an ingress is attached, the jenkins server node redirects the agents through the ingress address which only accepts connections through ports 80/443. This makes the agents unable to register itself to the server node.

By configuring agent nodes to use direction connection with tunneling, the agents can successfully connect.

### Rationale

Fixes bug where agents cannot connect to server node when ingress is applied.

### Juju Events Changes

None.

### Module Changes

`jenkins.py`: Creates a node definition first and instantiates the agent via modifying the tunnel property of the request.

### Library Changes

None.

### Checklist


src/agent.py
```diff
@@ -61,7 +61,10 @@ def _on_deprecated_agent_relation_joined(self, event: ops.RelationJoinedEvent) -
             event: The event fired from an agent joining the relationship.
         """
         container = self.charm.unit.get_container(JENKINS_SERVICE_NAME)
-        if not container.can_connect() or not self.state.is_storage_ready:
+        # This is to avoid the None type.
+        assert (binding := self.model.get_binding("juju-info"))  # nosec
```
yhaliaw, line 65: I am guessing the `get_binding` method will never return `None` here?
Can the comment be updated to include that? The current comment is a bit unclear.

src/agent.py
```diff
@@ -103,7 +105,10 @@ def _on_agent_relation_joined(self, event: ops.RelationJoinedEvent) -> None:
...
```
yhaliaw, line 109: Same as above.

yhaliaw (dismissed):

yhaliaw (approved):

PR #109, opened 2024-02-05
Support for auth_proxy integration

arturo-seijas:
Applicable spec: https://docs.google.com/document/d/1NCgjw7G6pT7qWe5EZ2IXoDqmPfN4NX214tZxQerHTmI

### Overview

Support for auth_proxy integration. For now, only authentication is supported, as the X-User header is the only one forwarded at the moment.

### Rationale

N/A

### Juju Events Changes

* Added auth_proxy_relation_created

### Module Changes

* Small refactoring of the jenkins module (permeates some other modules)
* New auth_proxy module

### Library Changes

* Added charms.oathkeeper.v0.auth_proxy

### Checklist


src/actions.py
```diff
@@ -49,7 +49,7 @@ def on_rotate_credentials(self, event: ops.ActionEvent) -> None:
             event: The rotate credentials event.
         """
         container = self.charm.unit.get_container(JENKINS_SERVICE_NAME)
-        if not container.can_connect() or not jenkins.is_storage_ready(container):
+        if not jenkins.is_storage_ready(container):
```
amandahla, line 52: I wonder if this couldn't be an Exception raised by rotate_credentials instead of checking this here.

src/auth_proxy.py
```diff
@@ -0,0 +1,61 @@
...
+            ingress: the ingress object from which to extract the necessary settings.
+        """
+        super().__init__(charm, "auth-proxy-observer")
+        self.charm = charm
+        self.ingress = ingress
+
+        self.auth_proxy = AuthProxyRequirer(self.charm)
+
+        self.charm.framework.observe(
+            self.charm.on["auth-proxy"].relation_joined, self._auth_proxy_relation_joined
```
amandahla, line 40: What happens for relation departed or changed?

src/jenkins.py
```diff
@@ -308,18 +310,37 @@ def _unlock_wizard(container: ops.Container) -> None:
...
+def _install_jenkins_config(container: ops.Container, filename: str) -> None:
     """Install jenkins-config.xml.
 
     Args:
         container: The Jenkins workload container.
+        filename: the source file to copy contents from.
     """
-    with open("templates/jenkins-config.xml", encoding="utf-8") as jenkins_config_file:
+    with open(filename, encoding="utf-8") as jenkins_config_file:
         container.push(CONFIG_FILE_PATH, jenkins_config_file, user=USER, group=GROUP)
```
amandahla, line 321: Shouldn't ops.pebble.PathError be handled here?

src/jenkins.py
```diff
@@ -308,18 +310,37 @@ def _unlock_wizard(container: ops.Container) -> None:
...
         container.push(LOGGING_CONFIG_PATH, jenkins_logging_config_file, user=USER, group=GROUP)
 
 
+def install_auth_proxy_config(container: ops.Container) -> None:
+    """Install jenkins-config.xml for auth_proxy.
+
+    Args:
+        container: The Jenkins workload container.
+    """
+    _install_jenkins_config(container, "templates/reverse-proxy-auth-jenkins-config.xml")
```
amandahla, line 341: Minor/nitpicking: could "templates/<file name>" be constants?

src/auth_proxy.py
arturo-seijas: Jenkins won't be receiving any data though this relation, it will just write it, so no need to react on relation changed or departed events

src/actions.py
arturo-seijas: Perhaps, but I'd rather keep the cahnges as contained as possible for this PR. This seems a bit out of the scope of the PR

src/jenkins.py
arturo-seijas: Updated

src/jenkins.py
arturo-seijas: Updated

src/auth_proxy.py
amandahla: Hm...So if you remove the relation after Jenkins is configured in install_auth_proxy_config, there is no need to revert anything right?

src/actions.py
amandahla: No worries, just caught my attention because the container.can_connect was moved to is_storage_ready.

src/auth_proxy.py
arturo-seijas: Mmm, you're right, we should revert the jenkins config change

arturo-seijas: Updated

.trivyignore
```diff
@@ -1,5 +1,7 @@
 # Jenkins CVEs
 CVE-2016-1000027
+CVE-2024-25710
```
yanksyoon, line 3: I think these CVEs can be removed (perhaps on another PR to help reduce complexity)

docs/how-to/integrate-with-iam.md
```diff
@@ -0,0 +1,37 @@
+# How to integrate with IAM
+
+This charm supports integration with the [IAM bundle](https://charmhub.io/iam) via [Oathkeeper](https://charmhub.io/oathkeeper), adding an authentication layer that will front the Jenkins applications. When enabled, the [reverse-proxy-auth-plugin](https://plugins.jenkins.io/reverse-proxy-auth-plugin/) will be installed and configured, blocking unaothorised access to Jenkins.
```
yanksyoon, line 3:
```suggestion
This charm supports integration with the [IAM bundle](https://charmhub.io/iam) via [Oathkeeper](https://charmhub.io/oathkeeper), adding an authentication layer that will front the Jenkins applications. When enabled, the [reverse-proxy-auth-plugin](https://plugins.jenkins.io/reverse-proxy-auth-plugin/) will be installed and configured, blocking unauthorised access to Jenkins.
```
just a small typo correction :D

docs/how-to/integrate-with-iam.md
```diff
@@ -0,0 +1,37 @@
...
+
+The steps to enable this mechanism are described below.
+
+## Deploy the IAM bundle
```
yanksyoon, line 7: I'm wondering if numerical headings could help since these are steps :)

docs/how-to/integrate-with-iam.md
```diff
@@ -0,0 +1,37 @@
...
+## Deploy Oathkeeper
+
+Oathkeeper will interface between Jenkins and the IAM bundle. You will need to deploy the charm and issue and configure TLS certificates for in-cluster communication. Note that the [self-signed-certificates charm](https://charmhub.io/self-signed-certificates) is already deployed as part of the IAM bundle.
+
+```
+juju deploy oathkeeper --channel edge --trust
+juju integrate oathkeeper:certificates self-signed-certificates
+```
+
+To leverage proxy authentication, enable traefiks feature and integrate the traefik charm instante with oathkeeper. As earlier, traefik-public is already deployed as part of the bundle.
```
yanksyoon, line 20:
```suggestion
To leverage proxy authentication, enable traefik's `enable_experimental_forward_auth` feature and integrate the traefik charm instance with Oathkeeper. As earlier, traefik-public is already deployed as part of the bundle.
```

requirements.txt
```diff
@@ -1,5 +1,6 @@
-cosl
-jenkinsapi>=0.3,<1
-ops>=2,<3
-pydantic>=1,<2
-requests>=2,<3
+cosl==0.0.7
```
yanksyoon, line 1: I'm wondering what the difference is in having pinned `requirements.txt` and `poetry.lock` might be.

src/actions.py
```diff
@@ -36,7 +36,7 @@ def on_get_admin_password(self, event: ops.ActionEvent) -> None:
             event: The event fired from get-admin-password action.
         """
         container = self.charm.unit.get_container(JENKINS_SERVICE_NAME)
```
yanksyoon, line 38: Out of the scope of this PR but I think this can now be refactored in charm init as per [this discourse post](https://discourse.charmhub.io/t/a-suggested-way-to-organise-the-code-inside-the-charm-py-file/13203/4) on organising `charm.py`

src/jenkins.py
```diff
@@ -288,47 +278,94 @@ def _unlock_wizard(container: ops.Container) -> None:
...
+        container: The Jenkins workload container.
+        filename: the source file to copy contents from.
+        destination_path: the target path.
+
+    Raises:
+        JenkinsBootstrapError: if the config can not be installed.
+
+    """
+    try:
+        with open(filename, encoding="utf-8") as jenkins_config_file:
```
yanksyoon, line 320: I think this can be made into `pathlib.Path` type and use `read_text()` for file contents to help remove the with clause :)

src/jenkins.py
```diff
@@ -288,47 +278,94 @@ def _unlock_wizard(container: ops.Container) -> None:
...
+
+    Raises:
+        JenkinsBootstrapError: if the token can not be setup.
     """
-    client = _get_client(get_admin_credentials(container))
-    token: str = client.generate_new_api_token("juju_api_token")
-    container.push(API_TOKEN_PATH, token, user=USER, group=GROUP)
+    try:
+        client = _get_client(get_admin_credentials(container))
+        token: str = client.generate_new_api_token("juju_api_token")
```
yanksyoon, line 365: I'm wondering if defining "juju_api_token" in a constant might help organize the code a little :)

templates/reverse-proxy-auth-jenkins-config.xml
```diff
@@ -0,0 +1,26 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<hudson>
+  <useSecurity>true</useSecurity>
+  <numExecutors>0</numExecutors>
+  <authorizationStrategy class="hudson.security.FullControlOnceLoggedInAuthorizationStrategy">
+    <denyAnonymousReadAccess>true</denyAnonymousReadAccess>
+  </authorizationStrategy>
+  <securityRealm class="org.jenkinsci.plugins.reverse_proxy_auth.ReverseProxySecurityRealm" plugin="reverse-proxy-auth-plugin@1.7.7">
+    <managerPasswordSecret>{AQAAABAAAAAQPt9fH9Ky04iMhNChDOcoYRX3maoyWTClSy2MTqXT850=}</managerPasswordSecret>
```
yanksyoon, line 9: I'm wondering if this is expected?

yanksyoon (commented):
Reviewed until templates/reverse-proxy-auth-jenkins-config.xml, commented minor questions, thank you!

src/actions.py
arturo-seijas: Let's keep that in mind

requirements.txt
arturo-seijas: We don't have peotly yet in this repo, but it will do the trick for pinning deps

docs/how-to/integrate-with-iam.md
arturo-seijas: There are some other analogous cases where this is the format that we follow, so I think consistency should be taken into account. We can discuss this for general guidance :)

templates/reverse-proxy-auth-jenkins-config.xml
arturo-seijas: Good catch. This is expected to be populated every time jenkins is restarted, so we can just remove the line :)

docs/how-to/integrate-with-iam.md
```diff
@@ -0,0 +1,37 @@
...
+juju integrate oathkeeper kratos
+```
+
+## Configure the Jenkins charm
+
+Jenkins needs to be accessible via the same ingress in which Oathkeeper has been configured for the requests to be redirected, so upon integrating with it and with Oathkeeper itself. Authentication is set up in place.
+```
+juju integrate jenkins-k8s:ingress traefik-public
+juju integrate oathkeeper jenkins-k8s:auth-proxy
+```
```
mthaddon, line 37: This seems to be missing a "and now you can connect to X URL and validate Y functionality".

Thanhphan1147 (dismissed):

src/auth_proxy.py
```diff
@@ -0,0 +1,86 @@
...
```
jdkandersson, line 40: A nitpick on this and below, I think the naming convention is `_on_<...>` for event handling functions?

src/auth_proxy.py
```diff
@@ -0,0 +1,86 @@
...
+            return
+
+        auth_proxy_config = AuthProxyConfig(
+            protected_urls=[self.ingress.url],
+            allowed_endpoints=AUTH_PROXY_ALLOWED_ENDPOINTS,
+            headers=AUTH_PROXY_HEADERS,
+        )
+        self.auth_proxy.update_auth_proxy_config(auth_proxy_config=auth_proxy_config)
+        # Jenkins Needs to be set up from scratch: requires config and env vars updates.
+        self.charm.on.jenkins_pebble_ready.emit(container)
```
jdkandersson, line 65: Perhaps, rather than emit an event, could we pull out the implementation of Jenkins pebble ready into a separate function, e.g. on the charm, and call it here?

src/auth_proxy.py
```diff
@@ -0,0 +1,86 @@
...
+        Args:
+            event: the event triggering the handler.
+        """
+        container = self.charm.unit.get_container(state.JENKINS_SERVICE_NAME)
+        if not jenkins.is_storage_ready(container):
+            logger.warning("Service not yet ready. Deferring.")
+            event.defer()  # The event needs to be handled after Jenkins has started(pebble ready).
+            return
+        # Jenkins Needs to be set up from scratch: requires config and env vars updates.
+        self.charm.on.jenkins_pebble_ready.emit(container)
```
jdkandersson, line 79: Similar as above

tests/integration/conftest.py
```diff
@@ -780,21 +778,212 @@ def external_hostname_fixture() -> str:
...
+@pytest.fixture()
+def external_user_email() -> str:
+    """Username for testing proxy authentication."""
+    return "admin@example.com"
+
+
+@pytest.fixture()
+def external_user_password() -> str:
+    """Password for testing proxy authentication."""
+    return "password"
```
jdkandersson, line 896: Should we do a randomised hex string here from the `secrets` module?

tests/integration/test_auth_proxy.py
```diff
@@ -0,0 +1,85 @@
+# Copyright 2024 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+"""Integration tests for jenkins-k8s-operator with auth_proxy."""
+
+# pylint: disable=unused-argument
```
jdkandersson, line 6: I think it might be possible to use a decorator to avoid needing this disable, see https://jdkandersson.com/2023/08/18/how-to-avoid-pylint-disableunused-argument-in-pytest/

tests/integration/test_auth_proxy.py
```diff
@@ -0,0 +1,85 @@
...
+    oathkeeper_related: Application,
+) -> None:
+    """
+    arrange: deploy the Jenkins charm and establish auth_proxy relations.
+    act: send a request Jenkins.
+    assert: a 401 is returned.
+    """
+    status = await model.get_status()
+    address = status["applications"]["traefik-public"]["public-address"]
+    response = requests.get(  # nosec
```
jdkandersson, line 33: What is the nosec here please?

tests/unit/test_auth_proxy.py
```diff
@@ -0,0 +1,165 @@
...
+    interface: auth_proxy
+  ingress:
+    interface: ingress
+containers:
+  jenkins:
+    resource: jenkins-image
+"""
+
+
+class TestCharm(ops.CharmBase):
```
jdkandersson, line 32: Could anycharm be used here?

jdkandersson (commented):
Thanks, I particularly like the integration test!

tests/unit/test_auth_proxy.py
arturo-seijas: It seems like a bit overkill since this is a unit tests and we don't really need the other side of the relation. This approach is also used in some other charms for unit testing libraries

tests/integration/test_auth_proxy.py
arturo-seijas: Added an inline comment with the explanation

src/auth_proxy.py
arturo-seijas: Indeed. Updated

tests/integration/test_auth_proxy.py
arturo-seijas: Nice! Thanks for the tip. Updated

src/auth_proxy.py
arturo-seijas: I see some issues with that, mainly that we'd be creating a cyclic dependency between this module and the charm module. I think we need to reassess our practices to discuss scenarios like this one

tests/integration/conftest.py
arturo-seijas: It would be a bit more complex that that, since I'd have to pass the bcript hashed password all the way down to the dex k8s descriptors. It is definitely possible, though, if you think that's worth it

weiiwang01 (approved):

src/auth_proxy.py
gregory-schiano: @weiiwang01 I think your comment is unnecessary, the problem here has nothing to do with Observer pattern, we should focus on solutions.

@arturo-seijas @jdkandersson
It's explicitly stated in the charm development best practices that this pattern must not be used (https://juju.is/docs/sdk/styleguide#heading--custom-events) and it has been confirmed in Riga during a meeting that included most charm developer leaders and many charm developers. Main reason is that it leads to non implicit code similar to what the reactive pattern was doing, and Ops philosophy is explicitly to avoid the implicit aspects of Reactive.

I don't see why we refactoring the code to extract the common logic required for both `pebble_ready` event and `auth_proxy` events is not possible.
The logic scoped to the `container` and `jenkins` service within the `pebble_ready` event hook method can be extracted into a separate module named `pebble` whose purpose is to manage the pebble plan (as a consequence all methods and constants linked to pebble layers and pebble plans must be moved to this module).
This module is either a class, and as a consequence takes the jenkins service as dependency
Or this module isn't a class and the jenkins service is passed along (becoming a dependency of the proxy observer).

Everything related to charm status or unit version stays in the hook methods (somehow duplicated code)

weiiwang01: I the it's relevant. If the observer pattern is not applied here, then there's no need for this discussion because there would be no problems. And, we will need no solutions since there are no problems to solve.

weiiwang01: I have no grudges against the observer pattern; I just have observations regarding its poor real-world performance and am unable to understand the strong preference on it.

The issue of the "calling pebble_ready hook" you mentioned does not exist elsewhere in the charm but is exclusive to the newly introduced observer. Therefore, the changes are unnecessary if no observers are present.

Also, the workarounds introduced to address problems invented by the observer pattern are far from beneficial. For instance, in `state.py`, it must access the relation data directly instead of using the charm library, due to the restriction of observer. In `pebble.py`, object-oriented programming is not feasible without introducing circular dependencies. (<- please don't ask for another refactor; I will approve the current state as is.)

More serious issues include the accessing relation data in `is_auth_proxy_integrated` in `state.py` before `_on_auth_proxy_relation_joined`, as relation data is set during `_on_auth_proxy_relation_joined`. This is another problem with charm state as well.

Handling the auth proxy relation is just a trival task without the observer. By simply running the same config change procedure during the auth proxy relation join and departure events, and in that procedure switch between with and without the auth proxy and it's done. That way is just more simple, easy, clean, compact, and less prone to error.

jdkandersson (approved):

Thanhphan1147 (dismissed):

src/state.py
```diff
@@ -185,6 +186,18 @@ def _get_agent_meta_map_from_relation(
...
+def _is_auth_proxy_integrated(relation: typing.Optional[ops.Relation]) -> bool:
+    """Check if there is an auth proxy integration..
+
+    Args:
+        relation: The auth-proxy relation.
+
+    Returns:
+        True if an integration for atuh proxy exists.
+    """
+    return bool(relation and relation.app and relation.data[relation.app])
```
weiiwang01, line 198: I think the data for the auth proxy relationship is configured on the requirer side and there's no data on the provider side https://github.com/canonical/charm-relation-interfaces/blob/main/interfaces/auth_proxy/v0/README.md. So, to verify the existence of data, `relation.data[self.app]` should be used rather than `relation.data[relation.app]`.

src/state.py
```diff
@@ -270,6 +285,9 @@ def from_charm(cls, charm: ops.CharmBase) -> "State":
             deprecated_agent_meta_map = _get_agent_meta_map_from_relation(
                 charm.model.relations[DEPRECATED_AGENT_RELATION], charm.app.name
             )
+            is_auth_proxy_integrated = _is_auth_proxy_integrated(
+                charm.model.get_relation(AUTH_PROXY_RELATION)
+            )
```
weiiwang01, line 290: There may be a problem here because the relational data is set during `_on_auth_proxy_relation_joined`. However, the function `from_charm` is called in `JenkinsK8sOperatorCharm.__init__`, which occurs before `_on_auth_proxy_relation_joined`.

src/state.py
arturo-seijas: I don't need to check if the relation data is present, since existence of the relation suffices so add authentication. changing it. Thanks for spotting this!

weiiwang01 (approved):

gregory-schiano (approved):
LGTM

PR #113, opened 2024-02-07
Add version upgrade test

gtrkiller:
Applicable spec: <link>

### Overview

Adds an integration test to verify that all the plugins are working once we update to a newer version of Jenkins.

### Rationale

Keep track of plugin functionality across versions

### Juju Events Changes

N/A

### Module Changes

N/A

### Library Changes

N/A

### Checklist

- [X] The [charm style guide](https://juju.is/docs/sdk/styleguide) was applied
- [X] The [contributing guide](https://github.com/canonical/is-charms-contributing-guide) was applied
- [X] The changes are compliant with [ISD054 - Managing Charm Complexity](https://discourse.charmhub.io/t/specification-isd014-managing-charm-complexity/11619)
- [X] The documentation is generated using `src-docs`
- [X] The documentation for charmhub is updated.


tests/integration/helpers.py
```diff
@@ -58,7 +58,7 @@ async def install_plugins(
     client.safe_restart()
     await unit.model.block_until(
         lambda: requests.get(web, timeout=10).status_code == 403,
-        timeout=300,
+        timeout=600,
```
arturo-seijas, line 61: 60 * 10?

tests/integration/test_upgrade.py
```diff
@@ -0,0 +1,107 @@
...
+async def test_jenkins_upgrade_check_plugins(  # pylint: disable=too-many-locals
+    ops_test: OpsTest, jenkins_image: str
+):
+    """
+    arrange: given charm has been built, deployed and plugins have been installed.
+    act: get Jenkins' version, upgrade the charm and if the versions differ, check plugins.
+    assert: all the installed plugins are up and running.
+    """
+    # Ignoring assignment lint error due to knowing there will always be a model.
+    model: Model = ops_test.model  # type: ignore[assignment]
```
arturo-seijas, line 72: I believe there's a fixture for `model`

tests/integration/test_upgrade.py
```diff
@@ -0,0 +1,107 @@
...
+
+
+async def test_jenkins_deploy_with_plugins(ops_test: OpsTest, jenkins_image: str):
+    """
+    arrange: given a juju model.
+    act: deploy Jenkins, instantiate the Jenkins client and install the plugins.
+    assert: the deployment has no errors.
+    """
+    # Ignoring assignment lint error due to knowing there will always be a model.
+    model: Model = ops_test.model  # type: ignore[assignment]
```
arturo-seijas, line 33: I believe there's a fixture for model

gtrkiller: there is, but as I need to use a totally diffferent model form that in the fixture, that deploys the local charm at first. this test requires a totally different module.

tests/integration/test_upgrade.py
```diff
@@ -0,0 +1,107 @@
...
+        "ssh",
+        "--container",
+        "jenkins",
+        jenkins_unit.name,
+        "cat",
+        str(jenkins.API_TOKEN_PATH),
+    )
+    assert ret == 0, f"Failed to get Jenkins API token, {stderr}"
+    jenkins_client = jenkinsapi.jenkins.Jenkins(
+        f"http://{address}:8080", "admin", api_token, timeout=600
```
arturo-seijas, line 55: Can you use `jenkins.WEB_PORT` instead for the port?

tests/integration/test_upgrade.py
```diff
@@ -0,0 +1,107 @@
...
+    )
+    unit_web_client = UnitWebClient(
+        unit=jenkins_unit, web=f"http://{address}:8080", client=jenkins_client
```
arturo-seijas, line 58: Same here

tests/integration/test_upgrade.py
```diff
@@ -0,0 +1,107 @@
...
+    )
+    await install_plugins(unit_web_client, ALL_PLUGINS)
+
+
+async def test_jenkins_upgrade_check_plugins(  # pylint: disable=too-many-locals
```
arturo-seijas, line 63: Can you check whether the above comments apply here?

tests/integration/test_upgrade.py
```diff
@@ -0,0 +1,107 @@
...
+    """
+    # Ignoring assignment lint error due to knowing there will always be a model.
+    model: Model = ops_test.model  # type: ignore[assignment]
+    application = model.applications["jenkins-k8s"]
+    status = await model.get_status()
+    unit = list(status.applications["jenkins-k8s"].units)[0]
+    address = status["applications"]["jenkins-k8s"]["units"][unit]["address"]
+    response = requests.get(f"http://{address}:8080", timeout=60)
+    version = response.headers["X-Jenkins"]
+    charm = await ops_test.build_charm(".")
```
arturo-seijas, line 79: There should be already a fixture for the charm

gtrkiller: See my comment for the model above, same applies here.

tests/integration/test_upgrade.py
gtrkiller: See my comment for the model above, same applies here.

tests/integration/test_upgrade.py
```diff
@@ -0,0 +1,107 @@
...
+    await application.refresh(path=charm, resources={"jenkins-image": jenkins_image})
+    await model.wait_for_idle(status="active", timeout=30 * 60)
+    status = await model.get_status()
+    unit = list(status.applications["jenkins-k8s"].units)[0]
```
arturo-seijas, line 83: Can you extract the lines to get the address to a helper method?

tests/integration/test_upgrade.py
gtrkiller: See my comment for the model above, same applies here.

tests/integration/test_upgrade.py
arturo-seijas: How come? Isn't it empty?
```
@pytest.fixture(scope="module", name="model")
def model_fixture(ops_test: OpsTest) -> Model:
    """The testing model."""
    assert ops_test.model
    return ops_test.model
```

tests/integration/test_upgrade.py
arturo-seijas: I was referring to using `jenkins.WEB_PORT` as the port number

tests/integration/test_upgrade.py
```diff
@@ -0,0 +1,99 @@
...
+    arrange: given a juju model.
+    act: deploy Jenkins, instantiate the Jenkins client and install the plugins.
+    assert: the deployment has no errors.
+    """
+    application: Application = await model.deploy(
+        "jenkins-k8s",
+        application_name=JENKINS_APP_NAME,
+        channel="edge",
+    )
+    await application.set_config({"allowed-plugins": ",".join(ALL_PLUGINS)})
```
arturo-seijas, line 39: This can be defined when running `model.deploy` if you specify the `config` attr

tests/integration/test_upgrade.py
```diff
@@ -0,0 +1,99 @@
...
+    await model.wait_for_idle(status="active", timeout=30 * 60)
+    address = await get_model_jenkins_unit_address(model, JENKINS_APP_NAME)
+    jenkins_unit = application.units[0]
+    ret, api_token, stderr = await ops_test.juju(
```
arturo-seijas, line 43: Can you extract this duplicated logic somewhere else?

tests/integration/test_upgrade.py
```diff
@@ -0,0 +1,99 @@
+# Copyright 2024 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+# pylint: disable=redefined-outer-name,unused-argument,duplicate-code
```
arturo-seijas, line 4: It would be nice to aim to remove the `duplicate-code` exclusion

tests/integration/helpers.py
```diff
@@ -267,6 +267,37 @@ async def wait_for(
...
+        "jenkins",
+        jenkins_unit.name,
+        "cat",
+        str(jenkins.API_TOKEN_PATH),
+    )
+    assert ret == 0, f"Failed to get Jenkins API token, {stderr}"
+    jenkins_client = jenkinsapi.jenkins.Jenkins(
+        f"http://{address}:{jenkins.WEB_PORT}", "admin", api_token, timeout=60 * 10
+    )
+    # pylint: enable=duplicate-code
```
arturo-seijas, line 294: Why do you need this pylint annotation?

tests/integration/helpers.py
```diff
@@ -267,6 +267,37 @@ async def wait_for(
...
+        "ssh",
+        "--container",
+        "jenkins",
+        jenkins_unit.name,
+        "cat",
+        str(jenkins.API_TOKEN_PATH),
+    )
+    assert ret == 0, f"Failed to get Jenkins API token, {stderr}"
+    jenkins_client = jenkinsapi.jenkins.Jenkins(
+        f"http://{address}:{jenkins.WEB_PORT}", "admin", api_token, timeout=60 * 10
```
arturo-seijas, line 292: You'll probably want to extract  `f"http://{address}:{jenkins.WEB_PORT}"` to a variable, since it's used twice

tests/integration/helpers.py
```diff
@@ -267,6 +267,37 @@ async def wait_for(
     raise TimeoutError()
 
 
+async def generate_client_from_application(ops_test, jenkins_unit, address) -> UnitWebClient:
```
arturo-seijas, line 270: Can you add type hints for the args?

arturo-seijas, line 270: You are actually generating it from the unit, not the application. Can you pass the application as an arg instead of the unit? Note that `jenkins_unit: Unit = application.units[0]` is always executed before invoking this method

arturo-seijas: You can also get the address `ops.test.model` and the application name, so you can drop the `address` arg

tests/integration/test_upgrade.py
```diff
@@ -0,0 +1,70 @@
...
+    generate_client_from_application,
+    get_model_jenkins_unit_address,
+    install_plugins,
+)
+
+LOGGER = logging.getLogger(__name__)
+JENKINS_APP_NAME = "jenkins-k8s-upgrade"
+
+
+async def test_jenkins_deploy_with_plugins(ops_test: OpsTest, model: Model):
```
arturo-seijas, line 29: Can you add a blank line to differentiate the arrange, act and assert blocks? https://github.com/canonical/is-charms-contributing-guide/blob/main/CONTRIBUTING.md#test-structure

tests/integration/helpers.py
```diff
@@ -251,6 +267,37 @@ async def wait_for(
...
+
+    Args:
+        ops_test: OpsTest framework
+        jenkins_unit: Juju Jenkins-k8s unit.
+        address: IP address from the Jenkins unit.
+
+    Returns:
+        A Jenkins web client.
+    """
+    # pylint: disable=duplicate-code
```
jdkandersson, line 281: What part of the code is this duplicated from please?

jdkandersson:
Just to clarify, the intent of the upgrade test is to make sure that upgrading to a new version of the charm with a new Jenkins LTS is smooth and user settings are retained etc. The current integration tests should cover that the plugins work. For example the test could be:

1. Deploy a previous revision of the charm, do some check that it works and apply at least 1 user setting
2. Deploy the new revision
3. Check that everything works and that settings are still as they were before

tests/integration/helpers.py
gtrkiller: The code here is the same used for the jenkins_client fixture in the conftest.py file

tests/integration/helpers.py
gtrkiller: The code here is similar to the one used for the jenkins_client fixture in the integration conftest.py file

tests/integration/helpers.py
gtrkiller: but the address var value changes between those uses, so it might not be a good idea

tox.ini
```diff
@@ -114,7 +114,7 @@ commands =
 description = Run integration tests
 deps =
     pytest
-    juju>=3,<4
+    juju<=3.3.0,>=3.0
```
arturo-seijas, line 117: I'd pin to juju==3.3.0.0, which is apparently working

tests/integration/helpers.py
```diff
@@ -251,6 +269,41 @@ async def wait_for(
...
+        ops_test: OpsTest framework
+        jenkins_app: Juju Jenkins-k8s application.
+
+    Returns:
+        A Jenkins web client.
+    """
+    assert ops_test.model
+    address = await get_model_jenkins_unit_address(ops_test.model, jenkins_app.name)
+    jenkins_unit = jenkins_app.units[0]
+    # pylint: disable=duplicate-code
```
arturo-seijas, line 287: I don't think you need to disable this anymore

tests/integration/helpers.py
```diff
@@ -251,6 +269,41 @@ async def wait_for(
...
+        "jenkins",
+        jenkins_unit.name,
+        "cat",
+        str(jenkins.API_TOKEN_PATH),
+    )
+    assert ret == 0, f"Failed to get Jenkins API token, {stderr}"
+    jenkins_client = jenkinsapi.jenkins.Jenkins(
+        f"http://{address}:{jenkins.WEB_PORT}", "admin", api_token, timeout=60 * 10
+    )
+    # pylint: enable=duplicate-code
```
arturo-seijas, line 300: Same here

tests/integration/helpers.py
gtrkiller: I've tested and it is necessary still :(

tests/integration/helpers.py
gtrkiller: See above

arturo-seijas (dismissed):

nrobinaubertin (dismissed):

tests/integration/test_upgrade.py
```diff
@@ -0,0 +1,67 @@
...
+    ops_test: OpsTest, jenkins_image: str, model: Model, charm: ops.CharmBase
+):
+    """
+    arrange: given charm has been built, deployed and a job has been defined.
+
+    act: get Jenkins' version and upgrade the charm.
+
+    assert: if Jenkins versions differ, the job persists.
+    """
+    application = model.applications[JENKINS_APP_NAME]
```
Thanhphan1147, line 56: nitpicking: you can also format this code to separate the arrange, act, and assert block

gtrkiller: They are already separated and no blank lines are accepted after the docstring.

tests/integration/test_upgrade.py
```diff
@@ -0,0 +1,67 @@
...
+
+async def test_jenkins_deploy_with_job(ops_test: OpsTest, model: Model):
+    """
+    arrange: given a juju model.
+
+    act: deploy Jenkins, instantiate the Jenkins client and define a job.
+
+    assert: the deployment has no errors.
+    """
+    application: Application = await model.deploy(
```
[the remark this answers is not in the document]
gtrkiller: They are already separated and no blank lines are accepted after the docstring.

tests/integration/helpers.py
jdkandersson: Perhaps this could be unified?

tests/integration/test_upgrade.py
```diff
@@ -10,48 +10,48 @@
...
-    address = await get_model_jenkins_unit_address(model, JENKINS_APP_NAME)
-    jenkins_unit = application.units[0]
-    unit_web_client = await generate_client_from_application(ops_test, jenkins_unit, address)
-    await install_plugins(unit_web_client, ALL_PLUGINS)
+    unit_web_client = await generate_client_from_application(ops_test, model, application)
+    unit_web_client.client.create_job(JOB_NAME, gen_git_test_job_xml("k8s"))
 
 
-async def test_jenkins_upgrade_check_plugins(
+async def test_jenkins_upgrade_check_job(
```
jdkandersson, line 46: It looks like this test depends on `test_jenkins_deploy_with_job` - perhaps it would be better to make it a fixture? Generally speaking tests should not depend on each other

jdkandersson (dismissed):

arturo-seijas (approved):

src/charm.py
```diff
@@ -188,16 +189,20 @@ def _remove_unlisted_plugins(self, container: ops.Container) -> ops.StatusBase:
...
         1. Remove plugins that are installed but are not allowed by plugins config value.
         2. Update Jenkins patch version if available and is within restart-time-range config value.
+
+        Args:
+            event: The update status event.
         """
         container = self.unit.get_container(JENKINS_SERVICE_NAME)
         if not jenkins.is_storage_ready(container):
             self.unit.status = ops.WaitingStatus("Waiting for container/storage.")
+            event.defer()
```
amandahla, line 205: The update status event occurs periodically at regular intervals..Wouldn't it be risky to defer it?

tests/integration/helpers.py
```diff
@@ -251,6 +269,54 @@ async def wait_for(
...
+    ret, api_token, stderr = await ops_test.juju(
+        "ssh",
+        "--container",
+        "jenkins",
+        jenkins_unit.name,
+        "cat",
+        str(jenkins.API_TOKEN_PATH),
+    )
+    assert ret == 0, f"Failed to get Jenkins API token, {stderr}"
+    return jenkinsapi.jenkins.Jenkins(f"{address}", "admin", api_token, timeout=60)
```
amandahla, line 295: Minor: Isn't address already a string?

tests/integration/test_upgrade.py
```diff
@@ -0,0 +1,73 @@
...
+
+    assert: if Jenkins versions differ, the job persists.
+    """
+    application = model.applications[JENKINS_APP_NAME]
+    unit_ip = await get_model_jenkins_unit_address(model, JENKINS_APP_NAME)
+    address = f"http://{unit_ip}:8080"
+    response = requests.get(address, timeout=60)
+    old_version = response.headers["X-Jenkins"]
+    await application.refresh(path=charm, resources={"jenkins-image": jenkins_image})
+    await model.wait_for_idle(status="active", timeout=30 * 60)
```
amandahla, line 64: Out of curiosity: Big timeout. Could it take that long?

src/charm.py
jdkandersson: Yeh I think in this case we should probably not defer

jdkandersson (dismissed):

jdkandersson (dismissed):

mthaddon (approved):

nrobinaubertin (approved):

PR #115, opened 2024-02-21
ISD-1622 Jenkins agent discovery with ingress

Thanhphan1147:
Applicable spec: ISD-126 (internal)

It's currently not possible for the [Jenkins charm](https://charmhub.io/jenkins-k8s) to establish a working cross-model relation with [machine agent charms ](https://charmhub.io/jenkins-agent)that does not have access to the k8s pod IP (only possible in microk8s) and with k8s agent charms that are not in the same cluster due to the Pod IP being hard-coded in the relation data bag.

This PR will add another ingress integration, called `agent-discovery-ingress`, which will be used for external agent discovery.

This PR will also enable websocket for all inbound agent connections. Using WebSocket, inbound agents can now be connected much more simply when a reverse proxy is present: if the HTTP(S) port is already serving traffic, most proxies will allow WebSocket connections with no additional configuration.

### Juju Events Changes

Modify `agent-relation-joined`

### Module Changes

Add observer for `agent-discovery-ingress`

### Checklist

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -18,17 +18,3 @@ options:
       Plugins installed by the user and their dependencies will be removed automatically if not on
       the list. Included plugins are not automatically installed.
     default: "bazaar,blueocean,dependency-check-jenkins-plugin,docker-build-publish,git,kubernetes,ldap,matrix-combinations-parameter,oic-auth,openid,pipeline-groovy-lib,postbuildscript,rebuild,reverse-proxy-auth-plugin,ssh-agent,thinBackup"
-  remoting-external-url:
-    type: string
-    description: >
-      Configure the charm to use this URL when establishing relations with agent charms.
-      This is useful when connecting agents from outside of the charm's Kubernetes cluster.
-      It is assumed that this url is reachable to the agents. A schema (http:// or https://)
-      is required
-    default: ""
-  remoting-enable-websocket:
-    type: boolean
-    description: >
-      Configure inbound agents to use Websocket and skip TCP port 50000.
-      This is useful when the charm is deployed behind a reverse-proxy or behind a firewall.
-    default: false
--- a/docs/how-to/integrate-with-external-agents.md
+++ b/docs/how-to/integrate-with-external-agents.md
@@ -17,4 +17,4 @@ juju integrate jenkins-k8s:agent <offer-endpoint>
 The charm assumes that:
 1. There are connectivity between the juju controller of the `jenkins-k8s` charm and the juju controller of the agent charm trying to connect with the `jenkins-k8s` charm.
 2. The agent can resolve the ingress hostname provided by the `jenkins-k8s` charm and the resulting IP address is reachable, and there are firewall rules in place to allow HTTP traffic.
-3. In case a reverse proxy is present, it is also expected that the HTTP connection coming from the agent charm is allowed to be upgraded into a Websocket connection.
+3. In case a reverse proxy is present, it is also expected that the HTTP connection coming from the agent charm is allowed to be upgraded into a Websocket connection. The reverse proxy should also be configured with a suitable idle timeout for websocket connections to avoid intermittent agent disconnection.
--- a/docs/how-to/integrate-with-external-agents.md
+++ b/docs/how-to/integrate-with-external-agents.md
@@ -1,0 +1,20 @@
+# How to integrate with external agent charms
+
+We consider any agent charm to be `external` when they don't have layer 3 connectivity with the `jenkins-k8s` charm. To integrate with those agent charms, we'll leverage the `jenkins-k8s` charm's `agent-discovery-ingress` integration.
+
+The `agent-discovery-ingress` integration can be used with any charm that supports the `:ingress` interface. One example is the [traefik-k8s](https://charmhub.io/traefik-k8s) charm.
+```bash
+juju integrate jenkins-k8s:agent-discovery-ingress traefik-k8s:ingress
+```
+
+Agents considered `external` have to be integrated using a cross-model integration. To integrate with such agent, simply integrate with the ingress provider charm as mentioned above and then integrate with the agent charm's offer endpoint.
+```bash
+juju integrate jenkins-k8s:agent-discovery-ingress traefik-k8s:ingress
+juju integrate jenkins-k8s:agent <offer-endpoint>
+```
+
+# Networking considerations
+The charm assumes that:
+1. There are connectivity between the juju controller of the `jenkins-k8s` charm and the juju controller of the agent charm trying to connect with the `jenkins-k8s` charm.
+2. The agent can resolve the ingress hostname provided by the `jenkins-k8s` charm and the resulting IP address is reachable, and there are firewall rules in place to allow HTTP traffic.
+3. In case a reverse proxy is present, it is also expected that the HTTP connection coming from the agent charm is allowed to be upgraded into a Websocket connection.
--- a/src/agent.py
+++ b/src/agent.py
@@ -79,7 +79,7 @@ class Observer(ops.Object):
         If we do not have an ingress, then use the pod ip as hostname.
         The reason to prefer this over the pod name (which is the actual
         hostname visible from the pod) or a K8s service, is that those
-        are routable virtually exclusively inside the cluster (as they rely)
+        are routable virtually exclusively inside the cluster as they rely
         on the cluster's DNS service, while the ip address is _sometimes_
         routable from the outside, e.g., when deploying on MicroK8s on Linux.
 
--- a/src/agent.py
+++ b/src/agent.py
@@ -3,10 +3,12 @@
 
 """The Jenkins agent relation observer."""
 import logging
+import socket
 import typing
 
 import ops
 
+import ingress
 import jenkins
 from state import AGENT_RELATION, DEPRECATED_AGENT_RELATION, JENKINS_SERVICE_NAME, AgentMeta, State
 
@@ -26,18 +28,24 @@ class AgentRelationData(typing.TypedDict):
 
 
 class Observer(ops.Object):
-    """The Jenkins agent relation observer."""
+    """The Jenkins agent relation observer.
 
-    def __init__(self, charm: ops.CharmBase, state: State):
+    Attributes:
+        agent_discovery_url: external hostname to be passed to agents for discovery.
+    """
+
+    def __init__(self, charm: ops.CharmBase, state: State, ingress_observer: ingress.Observer):
         """Initialize the observer and register event handlers.
 
         Args:
             charm: The parent charm to attach the observer to.
             state: The charm state.
+            ingress_observer: The ingress observer responsible for agent discovery.
         """
         super().__init__(charm, "agent-observer")
         self.charm = charm
         self.state = state
+        self.ingress_observer = ingress_observer
 
         charm.framework.observe(
             charm.on[DEPRECATED_AGENT_RELATION].relation_joined,
@@ -55,13 +63,40 @@ class Observer(ops.Object):
         )
         # Event hooks for agent-discovery-ingress
         charm.framework.observe(
-            charm.agent_discovery_ingress_observer.ingress.on.ready,
+            ingress_observer.ingress.on.ready,
             self.reconfigure_agent_discovery,
         )
         charm.framework.observe(
-            charm.agent_discovery_ingress_observer.ingress.on.revoked,
+            ingress_observer.ingress.on.revoked,
             self.reconfigure_agent_discovery,
         )
+
+    @property
+    def agent_discovery_url(self) -> str:
+        """Return the external hostname to be passed to agents via the integration.
+
+        If we do not have an ingress, then use the pod ip as hostname.
+        The reason to prefer this over the pod name (which is the actual
+        hostname visible from the pod) or a K8s service, is that those
+        are routable virtually exclusively inside the cluster (as they rely)
+        on the cluster's DNS service, while the ip address is _sometimes_
+        routable from the outside, e.g., when deploying on MicroK8s on Linux.
+
+        Attributes:
+            charm: The charm root JenkinsK8SOperatorCharm.
+        """
+        # Check if an ingress URL is available
+        try:
+            if ingress_url := self.ingress_observer.ingress.url:
+                return ingress_url
+        except ops.ModelError as e:
+            # We only log the error here as we can fallback to using pod IP
+            # if ingress is not available
+            logger.error(
+                "Failed obtaining agent discovery url: %s, is the charm shutting down?", e
+            )
+        # Fallback to pod IP
+        return f"http://{socket.getfqdn()}:{jenkins.WEB_PORT}"
 
     def _on_deprecated_agent_relation_joined(self, event: ops.RelationJoinedEvent) -> None:
         """Handle deprecated agent relation joined event.
@@ -94,7 +129,7 @@ class Observer(ops.Object):
             self.charm.unit.status = ops.BlockedStatus(f"Jenkins API exception. {exc=!r}")
             return
 
-        jenkins_url = self.state.agent_discovery_url
+        jenkins_url = self.agent_discovery_url
         event.relation.data[self.model.unit].update(
             AgentRelationData(url=jenkins_url, secret=secret)
         )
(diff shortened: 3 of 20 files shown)


config.yaml
```diff
@@ -18,17 +18,3 @@ options:
...
-      This is useful when connecting agents from outside of the charm's Kubernetes cluster.
-      It is assumed that this url is reachable to the agents. A schema (http:// or https://)
-      is required
-    default: ""
-  remoting-enable-websocket:
-    type: boolean
-    description: >
-      Configure inbound agents to use Websocket and skip TCP port 50000.
-      This is useful when the charm is deployed behind a reverse-proxy or behind a firewall.
-    default: false
```
cbartz, line 34: Just to be on the safe side, there is no deployment by anyone that relies on the existence of these configs, right?

src/agent.py
```diff
@@ -53,6 +62,65 @@ def __init__(self, charm: ops.CharmBase, state: State):
...
+        """Validate an IP address.
+
+        Args:
+            address: a string representing a unit address
+
+        Returns:
+            Whether the IP address is a valid unit IP address
+        """
+        try:
+            _ = ipaddress.ip_address(address)
```
cbartz, line 119:
```suggestion
            ipaddress.ip_address(address)
```
I think the assignment is not necessary.

tests/integration/test_ingress.py
```diff
@@ -32,3 +38,35 @@ async def test_ingress_integration(
...
+    await application.relate(AGENT_DISCOVERY_INGRESS_RELATION_NAME, traefik_application.name)
+
+    await model.wait_for_idle(
+        apps=[application.name, jenkins_k8s_agents.name, traefik_application.name],
+        wait_for_active=True,
+    )
+
+    agent_discovery_url_verified = False
+    for relation in application.relations:
+        if set([jenkins_k8s_agents.name, application.name]) == set(relation.applications):
```
cbartz, line 66:
```suggestion
        if {jenkins_k8s_agents.name, application.name} == set(relation.applications):
```
Consider this comment as optional

tests/integration/test_ingress.py
```diff
@@ -32,3 +38,35 @@ async def test_ingress_integration(
...
+    model: Model,
+    application: Application,
+    traefik_application: Application,
+    external_hostname: str,
+    jenkins_k8s_agents: Application,
+):
+    """
+    arrange: deploy the Jenkins charm and establish relations via ingress.
+    act: send a request to the ingress in /.
+    assert: the response succeeds.
```
cbartz, line 54: Does the test really do this? I actually see more of a relational data check.

tests/unit/test_agent.py
```diff
@@ -356,3 +361,177 @@ def test__on_agent_relation_departed(
...
+        "traefik-k8s",
+        app_data={"ingress": json.dumps({"url": mock_ingress_url})},
+    )
+
+    assert harness.charm.agent_observer.agent_discovery_url == mock_ingress_url
+
+
+def test_agent_discovery_url_fqdn_fallback(harness: Harness, monkeypatch: pytest.MonkeyPatch):
+    """
+    arrange: given a base jenkins charm with no ingress.
```
cbartz, line 387:
```suggestion
    arrange: given a base jenkins charm with no ingress and an invalid ip.
```

cbartz (commented):
Overall it looks good to me. Have a few comments. And I think you forgot to run src-docs.

tests/integration/test_ingress.py
Thanhphan1147: It doesn't actually, and the test was not very stable so I updated the test and this docstring.

config.yaml
Thanhphan1147: Not that I know of, and updating to a version without those config should be fine, as they only affect internal charm logic.

tests/integration/test_ingress.py
```diff
@@ -12,23 +12,77 @@
...
-    unit = next(iter(status.applications[application.name].units))
-    address = status["applications"][application.name]["units"][unit]["address"]
+    await application.relate("ingress", traefik_application.name)
+    await model.wait_for_idle(
+        apps=[application.name, traefik_application.name], wait_for_active=True, timeout=20 * 60
+    )
+    # Find traefik IP
+    status = await model.get_status(filters=[traefik_application.name])
+    unit = next(iter(status.applications[traefik_application.name].units))
+    traefik_address = status["applications"][traefik_application.name]["units"][unit]["address"]
```
cbartz, line 38: This seems to be duplicated below and could be extracted into a function.

cbartz (approved):
LGTM

tests/integration/test_ingress.py
Thanhphan1147: makes sense, I'll see if I can put it in the traefik charm fixture

Thanhphan1147: added it in the traefik charm's fixture

src/agent.py
```diff
@@ -53,6 +62,65 @@ def __init__(self, charm: ops.CharmBase, state: State):
...
         )
+        # Event hooks for agent-discovery-ingress
+        charm.framework.observe(
+            ingress_observer.ingress.on.ready,
+            self.reconfigure_agent_discovery,
+        )
+        charm.framework.observe(
+            ingress_observer.ingress.on.revoked,
+            self.reconfigure_agent_discovery,
+        )
```
jdkandersson, line 73: The guidance is that event handlers should be called `_on_<event name>`, is that applicable here please?

src/agent.py
```diff
@@ -53,6 +62,65 @@ def __init__(self, charm: ops.CharmBase, state: State):
...
+
+    @property
+    def agent_discovery_url(self) -> str:
+        """Return the external hostname to be passed to agents via the integration.
+
+        If we do not have an ingress, then use the pod ip as hostname.
+        The reason to prefer this over the pod name (which is the actual
+        hostname visible from the pod) or a K8s service, is that those
+        are routable virtually exclusively inside the cluster (as they rely)
```
jdkandersson, line 82: What is the reason for enclosing `as they rely` in brackets please?

src/agent.py
```diff
@@ -53,6 +62,65 @@ def __init__(self, charm: ops.CharmBase, state: State):
...
+        # Check if an ingress URL is available
+        try:
+            if ingress_url := self.ingress_observer.ingress.url:
+                return ingress_url
+        except ops.ModelError as exc:
+            # We only log the error here as we can fallback to using pod IP
+            # if ingress is not available
+            logger.error(
+                "Failed obtaining agent discovery url: %s, is the charm shutting down?", exc
+            )
```
jdkandersson, line 98: It is unusual to have an exception be raised on accessing a property? Perhaps getting the ingress URL should be a function? It also isn't clear why it is an `ops` error please?

src/agent.py
```diff
@@ -53,6 +62,65 @@ def __init__(self, charm: ops.CharmBase, state: State):
...
+
+        Returns:
+            Whether the IP address is a valid unit IP address
+        """
+        try:
+            ipaddress.ip_address(address)
+        except ValueError:
+            return False
+
+        return True
```
jdkandersson, line 123: Perhaps in-line this with a `pass` for when `ValueError` is raised?

src/ingress.py
```diff
@@ -12,12 +14,22 @@
 class Observer(ops.Object):
     """The Jenkins Ingress integration observer."""
 
-    def __init__(self, charm: ops.CharmBase):
+    def __init__(self, charm: ops.CharmBase, key: str, relation_name: typing.Optional[str] = None):
```
jdkandersson, line 17: What is the reason please for making `relation_name` optional?

jdkandersson: Couldn't we just always ask for it?

tests/integration/conftest.py
```diff
@@ -809,4 +809,8 @@ async def traefik_application_fixture(model: Model, external_hostname: str):
         raise_on_error=False,
     )
 
-    yield traefik
+    status = await model.get_status(filters=[traefik.name])
+    unit = next(iter(status.applications[traefik.name].units))
+    traefik_address = status["applications"][traefik.name]["units"][unit]["address"]
```
cbartz, line 814: I think I would have preferred to move this into a helper function, as this may not be necessary for all upcoming tests and we would not need to return a tuple here.

cbartz (approved):

src/agent.py
Thanhphan1147: The reason why I didn't name it `_on_<event_name>` is that the logic of handlers for those two events are exactly the same, so I was hesitating to have a method that does nothing other than calling another method, and if I implement those 2 methods separately they will basically have duplicated code, so in the end I just used `reconfigure_agent_discovery` as handler. Do you we should still have `on_<event_name>` hook that calls the `reconfigure_agent_discovery` method?

src/agent.py
Thanhphan1147: I didn't make much modification as this was taken from the [prometheus charm](https://github.com/canonical/prometheus-k8s-operator/blob/73cbee1c6ae5150d81bdef0682e67057ffbc6f9a/src/charm.py#L372), as far as I know, `ingress_observer.ingress.url` is a method with `@property` decorator, it also [fetch data from the model's relation data](https://github.com/canonical/traefik-k8s-operator/blob/cb25b7e4fb1f9d20c5a30fa1e8168a16812b1e7e/lib/charms/traefik_k8s/v2/ingress.py#L814) and it can raise `ops.ModelError` as a result

Thanhphan1147: I'll update the doc here to add more context

src/agent.py
Thanhphan1147: No particular reason as this was brought over from the prometheus charm and I didn't pay much attention to the docstring, I removed the brackets

src/agent.py
Thanhphan1147: I might need more context for why we'd want a `pass` when `ValueError` is raised here, my idea was that when `ValueError` is raised it means that the ip is not valid and in that case we should return `False` to fallback to using `socket.fqdn`

src/agent.py
jdkandersson: yes please 🙂

src/agent.py
jdkandersson: ah so what I was trying to get at is to remove this method and move it to the caller since that is a short function and it is only called by one method I think? That would simplify this a bit

src/agent.py
jdkandersson: Perhaps it also needs to be changed on the prometheus charm - what do you think about the feedback provided above?

tests/integration/test_ingress.py
```diff
@@ -6,29 +6,79 @@
...
+        f"http://{traefik_address}/{jenkins.LOGIN_URL}",
         headers={"Host": f"{model.name}-{application.name}.{external_hostname}"},
         timeout=5,
     )
-    assert response.status_code == 200
+
+    assert "Authentication required" in str(response.content)
+
+
+# This will only work on microk8s !!
```
jdkandersson, line 47: Is there a way to check this and fail the test if not on microk8s?

jdkandersson (commented):
hmm, perhaps I missed it, I didn't see a new integration test for the new integration?

cbartz (dismissed):

tests/integration/test_ingress.py
Thanhphan1147: This test is the test I added for the new integration. That's a good point as well, I'll see if I can add a check and output a meaningful error message when not running on microk8s

src/agent.py
Thanhphan1147: got it, I added those 2 handlers.

src/agent.py
Thanhphan1147: makes sense, I made this into a separate method since I expect we'd want to use this check elsewhere in our code eventually. I moved the method to its caller. If we want this IP check elsewhere in the charm in the future I'll move it back to a separate method.

src/agent.py
```diff
@@ -215,3 +241,30 @@ def _on_agent_relation_departed(self, event: ops.RelationDepartedEvent) -> None:
             self.charm.unit.status = ops.ActiveStatus(f"Failed to remove {agent_name}")
             return
         self.charm.unit.status = ops.ActiveStatus()
+
+    def _on_ready(self, event: IngressPerAppReadyEvent) -> None:
```
yanksyoon, line 245: I'm wondering if ingress related definitions can be prefixed with `ingress` to help avoid confusion to what is ready :)

src/ingress.py
```diff
@@ -12,12 +12,19 @@
 class Observer(ops.Object):
```
yanksyoon, line 12: I'm wondering if the `Observer` for ingress is strictly necessary - Could we instantiate IngressPerAppRequirer in agent observer or in charm class and pass it to agent observer?
My understanding is that an observer is used to simplify the `framework.observe` in the main charm class. Since this doesn't quite observer anything here, I'm wondering if removing this observer class would simplify the code. 
What do you think? :D

yanksyoon: Or we could move the agent_discovery_url and the observer events here so that the code related to ingress changes are more cohesive

docs/how-to/integrate-with-external-agents.md
```diff
@@ -0,0 +1,20 @@
...
+```bash
+juju integrate jenkins-k8s:agent-discovery-ingress traefik-k8s:ingress
+juju integrate jenkins-k8s:agent <offer-endpoint>
+```
+
+# Networking considerations
+The charm assumes that:
+1. There are connectivity between the juju controller of the `jenkins-k8s` charm and the juju controller of the agent charm trying to connect with the `jenkins-k8s` charm.
+2. The agent can resolve the ingress hostname provided by the `jenkins-k8s` charm and the resulting IP address is reachable, and there are firewall rules in place to allow HTTP traffic.
+3. In case a reverse proxy is present, it is also expected that the HTTP connection coming from the agent charm is allowed to be upgraded into a Websocket connection.
```
yanksyoon, line 20: I think we might need to be mindful of [websockets timing out](https://community.traefik.io/t/issue-with-application-using-websockets/18813/2), I've had issues with it before, not sure if it could apply here :)

yanksyoon (commented):
The docs are great and to the point! Minor comments added, thank you!

src/agent.py
Thanhphan1147: makes sense, I'm thinking of naming it `_ingress_on_ready` rather than `_on_ingress_ready` to stick with the naming convention

src/agent.py
Thanhphan1147: This implementation is also used in other projects as well and not just the prometheus charm I think. The decision to implementation fetching the ingress URL as a property comes from the traefik charm lib which is used by a lot of charms, so I think this change would be quite significant.

The ModelError raised from this is also something implemented by the traefik lib so I think a comment explaining why we catch that error here would be sufficient in the context of this PR.

docs/how-to/integrate-with-external-agents.md
Thanhphan1147: Good point, I haven't run into issues with websocket timing out yet, but I'll add it to the docs

src/agent.py
```diff
@@ -252,6 +242,22 @@ def _on_agent_relation_departed(self, event: ops.RelationDepartedEvent) -> None:
             return
         self.charm.unit.status = ops.ActiveStatus()
 
+    def _on_ready(self, event: IngressPerAppReadyEvent) -> None:
+        """Handle ready event for agent-discovery-ingress.
+
+        Args:
+            event: The event fired.
+        """
+        self.reconfigure_agent_discovery(event)
```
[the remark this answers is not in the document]
Thanhphan1147: Got it, I reordered those methods. Thanks for the tip.

jdkandersson (approved):

src/ingress.py
Thanhphan1147: Sounds reasonable, currently the agent observer needs to use the ingress Observer in its constructor to stick with the Architecture guidance and avoid cyclic dependency, so cleaning up the ingress observer would simplify this a bit. But merging the 2 observer does raise the question about separation of concerns. 

I'm marking this as resolved, let's continue the discussion and I'll open a follow-up PR if needed.

jdkandersson (approved):

cbartz (approved):

PR #120, opened 2024-03-07
Fix traefik path routing

arturo-seijas:
Applicable spec: <link> N/A

### Overview

Set a prefix for Jenkins when path routing is enabled. This means that the api calls will also need to take that prefix into account.

In order to achieve this, the prefix will be passed to the `jenkins.Jenkins` instance when constructed and the methods that ultimately make api calls are converted into instance methods

### Rationale

N/A

### Juju Events Changes

N/A

### Module Changes

Spreads across all modules

### Checklist


amandahla:
Is not necessary to observe the event _on_ingress_ready to get the URL when ready?

src/jenkins.py
```diff
@@ -144,6 +146,467 @@ def _get_api_credentials(container: ops.Container) -> Credentials:
...
+    @property
+    def login_url(self) -> str:
+        """Get the Jenkins login URL.
+
+        Returns: the login URL.
+        """
+        return f"{self.web_url}{LOGIN_PATH}"
+
+    @property
+    def version(self) -> str:
```
Thanhphan1147, line 187: Since we only expect the server version to change during a charm upgrade, should we just use a constant here?

Thanhphan1147 (commented):
only a minor comment, I haven't looked at tests yet

src/ingress.py
```diff
@@ -28,3 +30,16 @@ def __init__(self, charm: ops.CharmBase, key: str, relation_name: str):
             port=jenkins.WEB_PORT,
             strip_prefix=True,
         )
+
```
Thanhphan1147, line 33: should we add an integration test for this? (traefik in path routing mode)

tox.ini
```diff
@@ -125,10 +125,11 @@ deps =
...
     pytest --tb native --ignore={[vars]tst_path}unit --log-cli-level=INFO -s {posargs}
-    ; uncomment the following lines to use with output of tox -e build.
-    ; --jenkins-image=localhost:32000/jenkins:test \
-    ; --charm-file=./jenkins-k8s_ubuntu-22.04-amd64.charm \
-    ; --kube-config=~/.kube/config
+    # uncomment the following lines to use with output of tox -e build.
+    # --jenkins-image=localhost:32000/jenkins:test \
+    # --charm-file=./jenkins-k8s_ubuntu-22.04-amd64.charm \
+    # --kube-config=~/.kube/config \
+    # --keep-models
```
Thanhphan1147, line 132: I don't think we'd want to keep the test models by default right?

tests/unit/test_charm.py
```diff
@@ -125,67 +126,66 @@ def test__on_jenkins_pebble_ready_get_version_error(
     monkeypatch: pytest.MonkeyPatch,
 ):
     """
-    arrange: given a patched jenkins.get_version function that raises an exception.
+    arrange: given a patched jenkins.version function that raises an exception.
```
Thanhphan1147, line 129:
```suggestion
    arrange: given a patched jenkins.version property that raises an exception.
```

Thanhphan1147 (dismissed):
LGTM, only a few minor comments

weiiwang01 (dismissed):

src-docs/auth_proxy.py.md
jdkandersson, line 1: Should this file be here?

src/charm.py
```diff
@@ -5,6 +5,8 @@
 
 """Charm Jenkins."""
 
+# pylint: disable=too-many-instance-attributes
```
jdkandersson, line 8: Should this include a note as to why it is reasonable to disable this?

src/charm.py
```diff
@@ -99,14 +101,25 @@ def _get_pebble_layer(self, jenkins_env: jenkins.Environment) -> ops.pebble.Laye
...
         return ops.pebble.Layer(layer)
 
+    def calculate_env(self) -> jenkins.Environment:
+        """Return a dictionary for Jenkins Pebble layer.
+
+        Returns:
+            The dictionary mapping of environment variables for the Jenkins service.
+        """
+        return jenkins.Environment(
+            JENKINS_HOME=str(state.JENKINS_HOME_PATH),
```
jdkandersson, line 119: does this need to be on the state or could it just be a constant?

src/jenkins.py
```diff
@@ -3,6 +3,8 @@
 
 """Functions to operate Jenkins."""
 
+# pylint: disable=too-many-lines
```
jdkandersson, line 6: Perhaps this needs to be split into a few submodules now?

src/jenkins.py
```diff
@@ -144,6 +146,467 @@ def _get_api_credentials(container: ops.Container) -> Credentials:
         raise JenkinsBootstrapError("Admin API credentials not yet setup.") from exc
 
 
+class Jenkins:
```
jdkandersson, line 149: What is the reason please for changing this to be a class?

src/jenkins.py
```diff
@@ -144,6 +146,467 @@ def _get_api_credentials(container: ops.Container) -> Credentials:
...
+        """
+        self.environment = environment
+
+    @property
+    def web_url(self) -> str:
+        """Get the Jenkins web URL.
+
+        Returns: the web URL.
+        """
+        env_dict = typing.cast(typing.Dict[str, str], self.environment)
```
jdkandersson, line 175: Isn't this already declared as a dict? What is the reason for needing the cast here please?

jdkandersson (commented):
have reviewed up to the tests

src/ingress.py
arturo-seijas: I switched traefik's routing mode to `route` in the corresponding fixture. That should suffice (if route works, subdomain will work too)
```
traefik = await model.deploy(
        "traefik-k8s",
        channel="edge",
        trust=True,
        config={
            "external_hostname": external_hostname,
            "routing_mode": "path",
            "enable_experimental_forward_auth": True,
        },
    )
  
   ```

src/charm.py
arturo-seijas: It's a constant that is defined in the `state` module. I'll move it to `jenkins`

src/jenkins.py
arturo-seijas: Added a comment explaining why

src/jenkins.py
arturo-seijas: Yes, but it seems the me the PR is already too bit. I plan to do that in a follow up PR along with some more refactoring

src/charm.py
arturo-seijas: Removed. It's not needed anymore

src/jenkins.py
arturo-seijas: The client and all the method that use it are dependent on the prefix that is set in the pebble plan, as such, the easiest way to track that prefix is to set that prefix as an attribute of the `jenkins` class

src-docs/auth_proxy.py.md
arturo-seijas: Removed

src/jenkins.py
arturo-seijas: That would mean having the version defined in two places, the rockfile and the charm code. Anyway, this was already the existing behaviour, so I'd rather keep it as it is at least in this PR

weiiwang01 (approved):

Thanhphan1147 (approved):

weiiwang01 (approved):

Thanhphan1147 (approved):

mthaddon (approved):
LGTM, thx

PR #121, opened 2024-03-11
ISD-1657 add declarative pipeline plugin integration test

Thanhphan1147:
Add integration test for Pipeline:Declarative (pipeline-model-definition) plugin used to run declarative pipeline. 

### Checklist

### DIFF
--- a/tests/integration/helpers.py
+++ b/tests/integration/helpers.py
@@ -333,6 +333,30 @@ def kubernetes_test_pipeline_script() -> str:
     )
 
 
+def declarative_pipeline_script() -> str:
+    """Generate a declarative pipeline script.
+
+    Return:
+        The pipeline script
+    """
+    return textwrap.dedent(
+        """
+        pipeline {
+            agent any
+
+            stages {
+                stage('Integration Test') {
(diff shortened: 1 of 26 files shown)


tests/integration/helpers.py
```diff
@@ -333,6 +333,30 @@ def kubernetes_test_pipeline_script() -> str:
     )
 
 
+def declarative_pipeline_script() -> str:
```
arturo-seijas, line 336: I wonder if this should be a fixture instead

Thanhphan1147: I think I've had a discussion about this with someone, and it was decided that these should be functions instead of fixtures because we have a lot of them and making them a fixture would makes the `conftest.py` quite big and makes it quite complicated to track

arturo-seijas: Works for me

arturo-seijas (approved):

jdkandersson (approved):

amandahla (approved):
LGTM

jdkandersson (approved):

PR #131, opened 2024-03-22
Replan when ingress ready

Thanhphan1147:
When using traefik with path routing enabled, jenkins need to restart to update its prefix.

This PR also fix some issues with integration tests instability , as well as fix some bugs related to the identity-platform bundle used in `test_auth_proxy`.

Alos add a prefix when sending url for agent discovery.

### Checklist


jdkandersson (approved):

weiiwang01 (dismissed):

arturo-seijas (dismissed):

src/jenkins.py
```diff
@@ -1,3 +1,4 @@
+# pylint: disable=C0302
```
jdkandersson, line 1: What is this rule please? I think you can call it by name as well. Also, we usually ask for an explanatory comment as to why disabling is reasonable in this case please, do you think that is applicable here?

tests/integration/test_external_agent.py
```diff
@@ -0,0 +1,57 @@
...
+import typing
+
+import pytest
+from juju.application import Application
+
+import state
+from charm import AGENT_DISCOVERY_INGRESS_RELATION_NAME
+
+
+# This will only work on microk8s !!
```
jdkandersson, line 15: I can't remember, I think we somewhere have implemented a check to see if this is running on microk8s?

tests/unit/test_jenkins.py
```diff
@@ -5,7 +5,7 @@
 
 # Need access to protected functions for testing
 # All tests belong to single jenkins module
-# pylint:disable=protected-access, too-many-lines
+# pylint:disable=protected-access, C0302
```
jdkandersson, line 8: Similar here, referring to the rule by name makes it easier to know at least which one is disabled without needing to lookup a code 🙂

tests/integration/test_external_agent.py
yanksyoon: iirc, ops_test will automatically fail automatically :D

tests/integration/test_external_agent.py
```diff
@@ -0,0 +1,57 @@
...
+async def test_agent_discovery_ingress_integration(
+    application: Application,
+    traefik_application_and_unit_ip: typing.Tuple[Application, str],
+    external_hostname: str,
+    jenkins_machine_agents: Application,
+):
+    """
+    arrange: deploy the Jenkins charm, ingress, and a machine agent.
+    act: integrate the charms with each other and update the machine agent's dns record.
+    assert: All units should be in active status.
```
yanksyoon, line 26: I'm wondering if a check to test if the external agent is correctly registered would be beneficial here :D

yanksyoon (approved):
Out of curiosity, may I ask how the update prefix works? LGTM!

Thanhphan1147:
> Out of curiosity, may I ask how the update prefix works? LGTM!

The charm simply determines the prefix [here](https://github.com/canonical/jenkins-k8s-operator/blob/41829425f00ef8e0a62f9476e16446eb0d761957/src/charm.py#L62) during init by looking up the data that traefik sends over. Then the charm replans pebble for any ingress changes to apply the prefix here: https://github.com/canonical/jenkins-k8s-operator/blob/c0359b29e5f0e707c0c225c38218a7d6713f6be1/src/auth_proxy.py#L122-L131

src/jenkins.py
Thanhphan1147: I renamed this to its name. I think the name is giving enough context that we don't have to elaborate further.

tests/unit/test_jenkins.py
Thanhphan1147: I changed it back to using its name.

tests/integration/test_external_agent.py
Thanhphan1147: I just moved this test  from `test_ingress.py` to a new test file as sharing a traefik instance seemed too unstable (not from the charm's point of view but from ops_test's POV) and that module was failing quite frequently. Splitting that module into 2 test files seemed to have solved the instability issue.

This test is the only test where it only works on microk8s, and in its original PR [I mentioned that](https://github.com/canonical/jenkins-k8s-operator/pull/115/commits/2c00c97686066481356a17eec49732f0fb1e59ad) I couldn't find a good way to test for it, so I'll only keep a comment here for now.

tests/integration/test_external_agent.py
Thanhphan1147: I think in this test we actually run a job on the machine agent and check if the job passes, so I think that should suffice. what do you think?

yanksyoon: Thanks for letting me know!

src/agent.py
```diff
@@ -104,7 +104,8 @@ def agent_discovery_url(self) -> str:
             unit_ip = str(binding.network.bind_address)
             try:
                 ipaddress.ip_address(unit_ip)
-                return f"http://{unit_ip}:{jenkins.WEB_PORT}"
+                env_dict = typing.cast(typing.Dict[str, str], self.jenkins.environment)
```
yanksyoon, line 107: Very nitpicky (no need to implement here), perhaps the typing.cast can go under jenkins.Jenkins once so that others needing the environment can automatically infer it. Also perhaps a typed dict can help future developers know what the env contains :)

src/pebble.py
```diff
@@ -35,7 +39,7 @@ def replan_jenkins(
     try:
         jenkins_instance.wait_ready()
         # Tested in integration
-        if state.auth_proxy_integrated:  # pragma: no cover
+        if disable_security:  # pragma: no cover
```
yanksyoon, line 42: I'm wondering if a unit test would be worth it to remove the `no cover` statement even though it is tested in integration tests :) Also up to you!

tests/integration/conftest.py
```diff
@@ -3,6 +3,8 @@
 
 """Fixtures for Jenkins-k8s-operator charm integration tests."""
 
+# pylint: disable=too-many-lines
+
```
yanksyoon, line 7: Perhaps we should consider splitting the conftest or move the fixtures to according modules - if they're only used in one specific test.

tests/integration/test_auth_proxy.py
```diff
@@ -83,7 +83,8 @@ async def is_redirected_to_dex() -> Match[str] | None:
 
     # Choose provider
     async with page.expect_navigation():
-        await page.get_by_role("button", name="Dex").click()
+        # Increase timeout to wait for dex to be ready
+        await page.get_by_role("button", name="Dex").click(timeout=60000 * 5)
```
yanksyoon, line 87: Would timeout be better formatted in minutes i.e. 60 * number_of_seconds?

yanksyoon (dismissed):
LGTM! All nitpicks that can be out of scope, please feel free to resolve to merge :)

src/jenkins.py
```diff
@@ -285,6 +296,35 @@ def _setup_user_token(self, container: ops.Container) -> None:
...
+        except jenkinsapi.utils.requester.JenkinsAPIException as e:
+            # Jenkins api exception when generating user token
+            # We check if security is disabled
+            logger.info("Generate token failed, checking if security is disabled")
+            response = client.requester.get_url(
+                f"{client.base_server_url()}/manage/api/json?tree=useSecurity"
+            )
+            try:
+                if response.status_code == 200 and not response.json()["useSecurity"]:
+                    # !! Write a random string to the api token as a temporary workaround,
```
arturo-seijas, line 308: What are you working around with this? Why is this needed?

tests/unit/test_ingress.py
```diff
@@ -28,3 +32,54 @@ def test_get_path():
     assert harness.charm.ingress_observer.get_path() == ""
     ingress_per_app.url = None
     assert harness.charm.ingress_observer.get_path() == ""
+
+
+def test__on_ingress_replan(harness: Harness, monkeypatch: pytest.MonkeyPatch):
```
arturo-seijas, line 37: I think it's be better to name the test based on the actions executed and the expected result if that helps identify what the test is expected to do. For instance, for this case `test_traefik_integration_added_replans_jenkins`

tests/unit/test_jenkins.py
```diff
@@ -250,6 +251,87 @@ def test__setup_user_token_raises_exception(mock_env: jenkins.Environment):
...
 
 
+@pytest.mark.parametrize(
+    "exception",
+    [
+        pytest.param(requests.exceptions.JSONDecodeError, id="decode_error"),
+        pytest.param(KeyError, id="keyerror"),
+    ],
+)
+def test__setup_user_token_security_disabled_response_parse_error(
```
arturo-seijas, line 261: Note that ideally you should only test public APIs

src/jenkins.py
Thanhphan1147: yes, like we discussed before, the API token generation would need to be refactored so that the token is generated on-demand. But I thought it'll be too big of a change to implement it here so I added this as a temporary fix before doing a follow-up PR to refactor the token generation.

tests/unit/test_jenkins.py
Thanhphan1147: This test is only for the workaround that I mentioned above and should be removed with the refactor follow-up PR.

tests/unit/test_ingress.py
Thanhphan1147: good point, I'll rename this test method and take the opportunity to do some refactoring as well

Thanhphan1147: I renamed the tests in this module

tests/integration/test_auth_proxy.py
Thanhphan1147: I removed the timeout as it's no longer needed.

tests/integration/conftest.py
Thanhphan1147: Good point, I started with moving the auth_proxy related fixtures to the test module and removed this ignore line. I'll refactor more in a follow-up PR.

src/pebble.py
Thanhphan1147: I have to investigate further to understand why this was needed in the first place, I'll resolve this for now and do a follow-up PR to address this.

arturo-seijas (approved):

src/agent.py
Thanhphan1147: good point, I'll address this in a follow-up PR, alongside with the API token generation and tests refactoring

jdkandersson (approved):

PR #219, opened 2025-03-31
demo:uv+make transition

yanksyoon:
This is a demo branch!

To get-started, run `make`!

---

Any feedback all very welcome! Thank you!

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,111 @@
+PROJECT := $(dir $(abspath $(lastword $(MAKEFILE_LIST))))
+
+SRC := $(PROJECT)src
+TESTS := $(PROJECT)tests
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -0,0 +1,111 @@
...
+
+
+.PHONY: setup-development
+setup-development: ## Install development tools required for development
+	uv add bandit black codespell coverage isort mypy pydocstyle pylint pytest
+	uv add -r requirements.txt
+
+.PHONY: fmt
+fmt: ## Apply coding style standards to code
+	uv run isort $(ALL)
```
amandahla, line 24: What do you think about adding a check for uv installed? So user can be warned to run setup-development first.

PR #277, opened 2025-09-22
Add security policy

seb4stien:
Add security policy


erinecon (approved):
Thanks!
(Working on the linkcheck failures today, it's expected and not overly concerning, anyway it's not related to your PR so no action required on your part)

PR #379, opened 2026-06-08
feat(jcasc): add configuration-as-code to required plugins

yanksyoon:
## Summary
Add the `configuration-as-code` plugin to `REQUIRED_PLUGINS` so it is always installed during bootstrap. Also adds the `JCASC_CONFIG_PATH` constant pointing to `JENKINS_HOME/jenkins.yaml`.

This is the foundational change for JCasC support — all subsequent PRs in this stack depend on the plugin being available.

## Testing
Covered by existing bootstrap unit tests which verify all REQUIRED_PLUGINS are installed.

## Stack
- **PR1 (this) #379**: Plugin + constant ← main
- PR2 #380: Charm config option + state property + tests ← PR1
- PR3 #381: Pebble environment variables + test update ← PR2
- PR4 #382: Reconciliation logic + unit/integration tests ← PR3

## Ref
ISD-5501

PR #382, opened 2026-06-08
feat(jcasc): implement _reconcile_jcasc and JCasC lifecycle methods

yanksyoon:
## Summary
Implement the core JCasC reconciliation logic with a new **ReconcileBlockedError** pattern.

### Design: ReconcileBlockedError pattern
Sub-reconcilers now **raise `ReconcileBlockedError`** to signal blocked state instead of setting `self.unit.status` directly. The top-level `_reconcile()` catches it, sets `BlockedStatus`, and returns early — preventing downstream reconcilers from running and the unconditional `ActiveStatus` at the end from overwriting the blocked state.

### `_reconcile_jcasc()` in `charm.py`:
- Skips if Jenkins home dir is not ready (first boot before bootstrap)
- Parses user JCasC YAML; raises `ReconcileBlockedError` on invalid input
- Detects conflict between `auth_proxy` relation and `securityRealm` in config
- Injects admin credentials (\${JENKINS_ADMIN_PASSWORD}) if securityRealm not user-provided
- Writes config to container only when content changed (idempotent)
- Validates via `/configuration-as-code/check` endpoint
- Rolls back to previous config on validation failure
- Reloads via `/configuration-as-code/reload` (hot-reload, no restart)

### Jenkins class methods in `jenkins.py`:
- `reload_jcasc(container)`: POST to reload endpoint
- `check_jcasc(container, config_content)`: POST to check endpoint
- Both wrap all exceptions into `JenkinsError` for consistent error handling

### Unit tests (16 tests in `tests/unit/test_charm.py`):
- **TestReconcileJcascSkips**: Early exit on missing home dir, empty config, invalid YAML, non-dict YAML
- **TestReconcileJcascConflicts**: auth_proxy + securityRealm conflict detection; allowed without auth_proxy
- **TestReconcileJcascWrite**: Admin credential injection, idempotent no-write, make_dirs on push
- **TestReconcileJcascValidation**: Reload on success, rollback on failure, skip when Jenkins not ready
- **TestJenkinsJcascMethods**: reload_jcasc and check_jcasc method behavior

### Integration tests (4 tests in `tests/integration/test_jenkins.py`):
- Default config applied (export endpoint accessible)
- Custom config updates (systemMessage propagated)
- Invalid YAML blocks charm
- Hot-reload without restart

## Stack
- PR1 #379: Plugin + constant ← main
- PR2 #380: Charm config option + state property + unit tests ← PR1
- PR3 #381: Pebble environment variables + unit test update ← PR2
- **PR4 (this) #382**: Reconciliation logic + unit/integration tests ← PR3

## Ref
ISD-5501

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -178,9 +178,7 @@ class JenkinsK8sOperatorCharm(ops.CharmBase):
             return ""
         return path
 
-    def calculate_env(
-        self, admin_password: str = ""
-    ) -> jenkins.Environment:
+    def calculate_env(self, admin_password: str = "") -> jenkins.Environment:
         """Return a dictionary for Jenkins Pebble layer.
 
         Args:
@@ -268,6 +266,7 @@ class JenkinsK8sOperatorCharm(ops.CharmBase):
         if current_services != desired_services:
             container.add_layer(JENKINS_SERVICE_NAME, desired_layer, combine=True)
             container.replan()
+            self.jenkins.wait_ready()
 
     def _reconcile_jcasc(self, container: ops.Container, state: State) -> None:
         """Reconcile JCasC configuration to desired state.
@@ -287,9 +286,7 @@ class JenkinsK8sOperatorCharm(ops.CharmBase):
         desired_yaml = yaml.dump(desired_config, default_flow_style=False, sort_keys=False)
 
         if not self.jenkins.sync_jcasc_config(container, desired_yaml):
-            raise ReconcileBlockedError(
-                "JCasC validation failed — check juju debug-log"
-            )
+            raise ReconcileBlockedError("JCasC validation failed — check juju debug-log")
 
     def _build_jcasc_config(self, state: State) -> typing.Dict[str, typing.Any]:
         """Build the desired JCasC config by merging user config with charm-managed sections.
--- a/src/charm.py
+++ b/src/charm.py
@@ -227,9 +227,7 @@ class JenkinsK8sOperatorCharm(ops.CharmBase):
             self._reconcile_agents(event, state)
             self._reconcile_agent_discovery()
             self._reconcile_auth_proxy(event, state)
-            # Plugin removal only runs on update-status (matching original behaviour)
-            if isinstance(event, ops.UpdateStatusEvent):
-                self._reconcile_plugins(container, state)
+            self._reconcile_plugins(container, state)
         except ReconcileBlockedError as exc:
             self.unit.status = ops.BlockedStatus(exc.message)
             return
@@ -305,7 +303,6 @@ class JenkinsK8sOperatorCharm(ops.CharmBase):
         Returns:
             The merged JCasC configuration dict.
         """
-        # state.jcasc_config is already validated as a dict by State.from_charm
         config: typing.Dict[str, typing.Any] = dict(state.jcasc_config)  # type: ignore[arg-type]
         jenkins_section: typing.Dict[str, typing.Any] = config.get("jenkins", {})
 
--- a/src/charm.py
+++ b/src/charm.py
@@ -14,6 +14,7 @@ import typing
 from urllib.parse import urlparse
 
 import ops
+import yaml
 from charms.grafana_k8s.v0.grafana_dashboard import GrafanaDashboardProvider
 from charms.loki_k8s.v0.loki_push_api import LogProxyConsumer
 from charms.oauth2_proxy_k8s.v0.auth_proxy import AuthProxyConfig, AuthProxyRequirer
@@ -39,6 +40,23 @@ from state import (
 )
 
 logger = logging.getLogger(__name__)
+
+
+class ReconcileBlockedError(Exception):
+    """Raised by sub-reconcilers to signal the charm should enter BlockedStatus.
+
+    Attributes:
+        message: The blocked status message.
+    """
+
+    def __init__(self, message: str):
+        """Initialize ReconcileBlockedError.
+
+        Args:
+            message: The blocked status message to surface to the user.
+        """
+        self.message = message
+        super().__init__(message)
 
 
 class JenkinsK8sOperatorCharm(ops.CharmBase):
@@ -199,17 +217,22 @@ class JenkinsK8sOperatorCharm(ops.CharmBase):
         if state is None:
             return
 
-        # Storage ownership only needs correction on attach/upgrade events
-        if isinstance(event, (ops.StorageAttachedEvent, ops.UpgradeCharmEvent)):
-            self._reconcile_storage(container)
+        try:
+            # Storage ownership only needs correction on attach/upgrade events
+            if isinstance(event, (ops.StorageAttachedEvent, ops.UpgradeCharmEvent)):
+                self._reconcile_storage(container)
 
-        self._reconcile_pebble(container, state)
-        self._reconcile_agents(event, state)
-        self._reconcile_agent_discovery()
-        self._reconcile_auth_proxy(event, state)
-        # Plugin removal only runs on update-status (matching original behaviour)
-        if isinstance(event, ops.UpdateStatusEvent):
-            self._reconcile_plugins(container, state)
+            self._reconcile_pebble(container, state)
+            self._reconcile_jcasc(container, state)
+            self._reconcile_agents(event, state)
+            self._reconcile_agent_discovery()
+            self._reconcile_auth_proxy(event, state)
+            # Plugin removal only runs on update-status (matching original behaviour)
+            if isinstance(event, ops.UpdateStatusEvent):
+                self._reconcile_plugins(container, state)
+        except ReconcileBlockedError as exc:
+            self.unit.status = ops.BlockedStatus(exc.message)
+            return
 
         self.unit.status = ops.ActiveStatus(self._agent_status_message)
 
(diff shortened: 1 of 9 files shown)


src/jenkins.py
```diff
@@ -736,6 +736,123 @@ def plugins = jenkins.model.Jenkins.instance.getPluginManager().getPlugins()
...
+                        make_dirs=True,
+                    )
+                    self.reload_jcasc(container)
+                else:
+                    container.remove_path(jcasc_path)
+                return False
+
+            self.reload_jcasc(container)
+        except JenkinsBootstrapError:
+            logger.debug("Jenkins not yet bootstrapped, skipping validation/reload")
```
m2-farzan, line 849: We restart pebble layer in the `_reconcile_pebble`. I think it make it likely that jenkins will be booting at this point so this error is too likely to happen. Can we wait on jenkins to become ready (maybe with a timeout or backoff) to make the charm not have to wait for another event?

m2-farzan: Maybe use `self.jenkins.wait_ready()`

m2-farzan (approved):

src/jenkins.py
yanksyoon: It seems like then we should wait for the _reconcile_pebble to make Jenkins ready.

yanksyoon: Updated to add wait_ready condition in reconcile_pebble step!

florentianayuwono (approved):
lgtm thanks yanks

src/jenkins.py
```diff
@@ -736,6 +736,120 @@ def plugins = jenkins.model.Jenkins.instance.getPluginManager().getPlugins()
...
+            container: The Jenkins workload container.
+
+        Raises:
+            JenkinsError: if the reload request fails.
+        """
+        try:
+            credentials = _get_api_credentials(container)
+            client = self._get_client(credentials)
+            client.requester.post_url(f"{self.web_url}/configuration-as-code/reload")
+        except JenkinsError:
```
yhaliaw, line 752: 🟠 **Major — no-op `except JenkinsError: raise`**

This clause does nothing: Python propagates uncaught exceptions naturally. Its presence is misleading — a reader assumes it performs some handling — and it's dead code.

**Fix:** Delete the entire `except JenkinsError: raise` block; the surrounding `try` only needs to handle the non-`JenkinsError` case.

src/jenkins.py
```diff
@@ -736,6 +736,120 @@ def plugins = jenkins.model.Jenkins.instance.getPluginManager().getPlugins()
...
+        """
+        try:
+            credentials = _get_api_credentials(container)
+            client = self._get_client(credentials)
+            response = client.requester.post_url(
+                f"{self.web_url}/configuration-as-code/check",
+                data=config_content,
+            )
+            return response.status_code == 200
+        except JenkinsError:
```
yhaliaw, line 782: 🟠 **Major — same no-op `except JenkinsError: raise` as in `reload_jcasc`**

Same issue as above — delete this clause.

src/charm.py
```diff
@@ -225,6 +244,61 @@ def _reconcile_pebble(self, container: ops.Container, charm_state: State) -> Non
...
+            container: The Jenkins workload container.
+            state: The current charm state.
+        """
+        if state.jcasc_config is None:
+            return
+
+        desired_config = self._build_jcasc_config(state)
+        desired_yaml = yaml.dump(desired_config, default_flow_style=False, sort_keys=False)
+
+        if not self.jenkins.sync_jcasc_config(container, desired_yaml):
```
yhaliaw, line 266: 🟠 **Major — `JenkinsError` from `sync_jcasc_config` is unhandled**

Per the `sync_jcasc_config` docstring, `JenkinsError` propagates when Jenkins API calls fail. `_reconcile_jcasc` does not catch it, and `_reconcile` (the caller) only handles `ReconcileBlockedError`. An API failure here will propagate as an unhandled exception to the ops framework, setting the unit to **error status** rather than **blocked** — bypassing the consistent error-reporting pattern this PR establishes.

**Fix:** Catch `JenkinsError` here and re-raise as `ReconcileBlockedError`, or add a `JenkinsError` handler in `_reconcile`.

yhaliaw (commented):
I generated the review with an agent. I removed the ones I think are wrong, but left some comments that I am unsure about. Feel free to resolve the conversion.

Trying this out on this PR since it already has two approvals.

## Code Review: JCasC Reconcile Lifecycle

**Total findings:** 15 (🔴 Critical: 1 | 🟠 Major: 4 | 🟡 Minor: 6 | 🔵 Suggestions: 4)

The design — raising `ReconcileBlockedError` to halt the reconcile pipeline — is clean and consistent. The most serious issue is a silent error-swallowing bug in `sync_jcasc_config` (see inline). See also `src/state.py`, `tests/unit/test_state.py`, `tests/unit/test_ingress.py`, `tests/integration/helpers.py`, and `tox.toml`: no issues found in those files.

tests/unit/test_charm.py
```diff
@@ -537,10 +580,356 @@ def test__auth_proxy_relation_handlers_delegate(
...
     """
     harness_container.harness.begin()
-    jenkins_charm = typing.cast(JenkinsK8sOperatorCharm, harness_container.harness.charm)
+    jenkins_charm = typing.cast(
+        JenkinsK8sOperatorCharm, harness_container.harness.charm
+    )
     event = MagicMock(spec=event_type)
 
     with patch.object(jenkins_charm, "_reconcile_auth_proxy") as reconcile_auth_mock:
         jenkins_charm._reconcile(event)
```
yhaliaw, line 589: 🟠 **Major — deleted assertion makes this test vacuous**

The test patches `_reconcile_auth_proxy` as `reconcile_auth_mock` but never asserts that mock was called. It will pass even if `_reconcile_auth_proxy` is removed from `_reconcile` entirely.

**Fix:** Add `reconcile_auth_mock.assert_called_once()` after the `with` block.

tests/unit/test_charm.py
```diff
@@ -537,10 +580,356 @@ def test__auth_proxy_relation_handlers_delegate(
...
+            "local": {
+                "allowsSignup": False,
+                "users": [{"id": "custom", "password": "secret"}],
+            }
+        }
+    }
+}
+
+
+@pytest.fixture(name="harness_with_jcasc")
```
yhaliaw, line 611: 🟡 **Minor — unused fixture**

`harness_with_jcasc` is defined here but not referenced by any test in this file. Either remove it or add the missing consuming test.

tests/unit/test_charm.py
```diff
@@ -537,10 +580,356 @@ def test__auth_proxy_relation_handlers_delegate(
...
+
+    with patch.object(charm.jenkins, "sync_jcasc_config", return_value=sync_return):
+        if expect_blocked:
+            with pytest.raises(ReconcileBlockedError, match="validation failed"):
+                charm._reconcile_jcasc(mock_container, mock_state)
+        else:
+            charm._reconcile_jcasc(mock_container, mock_state)
+
+
+def test_sync_jcasc_config_no_write_when_unchanged():
```
yhaliaw, line 742: 🟡 **Minor — docstring does not follow the required AAA format**

Suggested replacement:
```python
"""
Arrange: given container already has the desired config (pull returns desired yaml).
Act: when sync_jcasc_config is called with the same content.
Assert: push is NOT called and True is returned.
"""
```

tests/unit/test_charm.py
```diff
@@ -537,10 +580,356 @@ def test__auth_proxy_relation_handlers_delegate(
...
+    mock_pull.read.return_value = desired
+    mock_container.pull.return_value = mock_pull
+
+    result = j.sync_jcasc_config(mock_container, desired)
+
+    assert result is True
+    mock_container.push.assert_not_called()
+
+
+def test_sync_jcasc_config_successful_write():
```
yhaliaw, line 761: 🟡 **Minor — docstring does not follow the required AAA format**

Suggested replacement:
```python
"""
Arrange: given first-time write (file doesn't exist) and check_jcasc returns True.
Act: when sync_jcasc_config is called with new content.
Assert: push uses make_dirs=True, reload_jcasc is called once, returns True.
"""
```

tests/unit/test_charm.py
```diff
@@ -537,10 +580,356 @@ def test__auth_proxy_relation_handlers_delegate(
...
+    if has_previous:
+        assert mock_container.push.call_count == 2
+        rollback_call = mock_container.push.call_args_list[1]
+        assert rollback_call[0][1] == previous
+        reload_mock.assert_called_once_with(mock_container)
+    else:
+        mock_container.remove_path.assert_called_once()
+
+
+def test_sync_jcasc_config_skips_validation_when_not_bootstrapped():
```
yhaliaw, line 825: 🔵 **Suggestion — this test would also pass for a plain `JenkinsError`, masking the Critical bug**

The production `sync_jcasc_config` currently catches the broader `JenkinsError`. This test patches `check_jcasc` to raise `JenkinsBootstrapError`, but since `JenkinsBootstrapError` is a subclass of `JenkinsError`, the test passes whether the production catch is `JenkinsBootstrapError` or `JenkinsError`.

After fixing the Critical finding (tightening the catch), add a complementary test: patch `check_jcasc` to raise a plain `JenkinsError` and verify it **propagates** (is not swallowed). This would have caught the Critical bug in CI.

tests/unit/test_charm.py
```diff
@@ -537,10 +580,356 @@ def test__auth_proxy_relation_handlers_delegate(
...
+        patch.object(j, "reload_jcasc") as reload_mock,
+    ):
+        result = j.sync_jcasc_config(mock_container, "jenkins:\n  test: true\n")
+
+    assert result is True
+    mock_container.push.assert_called_once()
+    reload_mock.assert_not_called()
+
+
+def test_reload_jcasc_success():
```
yhaliaw, line 851: 🟡 **Minor — docstring does not follow the required AAA format**

Suggested replacement:
```python
"""
Arrange: given a Jenkins instance with a mocked client that returns successfully.
Act: when reload_jcasc is called.
Assert: POST to /configuration-as-code/reload is made exactly once.
"""
```

tests/unit/test_charm.py
```diff
@@ -537,10 +580,356 @@ def test__auth_proxy_relation_handlers_delegate(
...
+    ):
+        creds_mock.return_value = jenkins.Credentials("admin", "token")
+        j.reload_jcasc(mock_container)
+
+    mock_requester.post_url.assert_called_once_with(
+        "http://localhost:8080/configuration-as-code/reload"
+    )
+
+
+def test_reload_jcasc_raises_jenkins_error():
```
yhaliaw, line 875: 🟡 **Minor — docstring does not follow the required AAA format**

Suggested replacement:
```python
"""
Arrange: given _get_api_credentials raises ConnectionError.
Act: when reload_jcasc is called.
Assert: JenkinsError is raised.
"""
```

src/jenkins.py
yhaliaw: Is this to add the error entry to the docstring?
Can a short code comment be added if that is the case?

PR #389, opened 2026-06-23
test: refactor pebble tests

yanksyoon:
Stack PR 1/8

Head: `chore/refactor-pebble-unit-tests`
Base: `feature/jcasc-reconcile`

This PR is part of a stacked sequence; review/merge in order.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -165,12 +165,12 @@ class JenkinsK8sOperatorCharm(ops.CharmBase):
             return ""
         return path
 
-    def calculate_env(self, config_hash: str, admin_password: str = "") -> jenkins.Environment:
+    def calculate_env(self, config_hash: str, admin_password: str) -> jenkins.Environment:
         """Return a dictionary for Jenkins Pebble layer.
 
         Args:
+            config_hash: The hash of the JCasC configurations applied.
             admin_password: The admin password for JCasC secret interpolation.
-                Empty string if not yet available (e.g. during __init__).
 
         Returns:
             The dictionary mapping of environment variables for the Jenkins service.
--- a/src/charm.py
+++ b/src/charm.py
@@ -131,12 +131,8 @@ class JenkinsK8sOperatorCharm(ops.CharmBase):
             self.on.update_status,
         ]:
             self.framework.observe(event, self._reconcile)
-        self.framework.observe(
-            self.on.get_admin_password_action, self._on_get_admin_password
-        )
-        self.framework.observe(
-            self.on.rotate_credentials_action, self._on_rotate_credentials
-        )
+        self.framework.observe(self.on.get_admin_password_action, self._on_get_admin_password)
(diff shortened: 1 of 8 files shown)


DeeKay3 (approved):

src/charm.py
```diff
@@ -218,120 +246,98 @@ def _reconcile_storage(self, container: ops.Container) -> None:
...
 
-        if not self.jenkins.sync_jcasc_config(container, desired_yaml):
-            raise ReconcileBlockedError("JCasC validation failed — check juju debug-log")
+        # Backwards compatibility: fetch admin password from container and migrate to juju secrets.
+        admin_setup = False
+        try:
+            password_or_token = jenkins.get_admin_credentials(container).password_or_token
+            admin_setup = True
+        except jenkins.JenkinsBootstrapError as exc:
+            logger.debug("Admin password not yet setup, setting up admin user: %s", exc)
```
yhaliaw, line 269: In this error path, for `get_admin_credentials` to raise `JenkinsBootstrapError`?
That should result in `password_or_token` to fail to assign a value.
If `not admin_setup`, then None will be returned at line 276.
But the type hints in this function is `str` for return value?

Can you check if this is possible?

src/charm.py
```diff
@@ -218,120 +246,98 @@ def _reconcile_storage(self, container: ops.Container) -> None:
...
 
-    def _build_jcasc_config(self, state: State) -> typing.Dict[str, typing.Any]:
-        """Build the desired JCasC config by merging user config with charm-managed sections.
+        # Generate admin password and set it using juju secrets if secret not yet configured.
+        if not admin_setup:
+            # Generate admin user secret using secrets.token_hex() and set it in the container
+            password_or_token = secrets.token_hex(16)
+        self.app.add_secret(content={"password": password_or_token}, label=self.app.name)
```
yhaliaw, line 275: Should this be only under the `if not admin_setup:` clause?
Seems weird to add secret if the value is not changed.

src/charm.py
```diff
@@ -364,25 +369,40 @@ def _reconcile_auth_proxy(self, event: ops.EventBase, state: State) -> None:
                 )
             self._auth_proxy.update_auth_proxy_config(auth_proxy_config=auth_proxy_config)
 
-    def _reconcile_plugins(self, container: ops.Container, state: State) -> None:
+    def _reconcile_plugins(
+        self, state: State, admin_client: jenkins.Jenkins, container: ops.Container
```
yhaliaw, line 373: It seems the `test_charm.py` is still using the old function sig: https://github.com/canonical/jenkins-k8s-operator/blob/cdfe3b8947342d21a4aa1e9971f058a331cbab34/tests/unit/test_charm.py#L198
There seems to be 4 instances of usage in that test file.

yhaliaw: btw, would not have caught this without AI.

PR #406, opened 2026-07-02
feat(jcasc): add jcasc-repository config option for git-based config

yanksyoon:
Applicable spec: <link>

### Overview

Implements the jcasc-repository feature to allow cloning Jenkins Configuration as Code from a git repository, with optional token-based authentication for private repos.

### Rationale

To allow git repository management of jenkins

### Checklist

- [x] The docs/changelog.md is updated with user-relevant changes in the format of [keep a changelog v1.1.0](https://keepachangelog.com/en/1.1.0/)


src/jenkins.py
```diff
@@ -1204,3 +1206,165 @@ def sync_jcasc_config(container: ops.Container, configuration_yaml: str) -> str:
...
+            logger.info(
+                "Cloning JCasC repository: url=%s branch=%s config_path=%s dest=%s",
+                url,
+                branch,
+                config_path,
+                dest,
+            )
+            container.exec(clone_command, environment=environment, timeout=300).wait_output()
+            logger.info("Repository clone successful: %s", url)
+        except ops.pebble.ExecError as exc:
```
yhaliaw: Only `ops.pebble.ExecError` is caught here. Every other spot in this file that waits on a container exec and needs to be resilient (`is_storage_ready` at line 728, and line 1077) catches `(ops.pebble.ChangeError, ops.pebble.ExecError)` together. If the clone exec raises `ChangeError` instead of `ExecError`, it will propagate out of `fetch_jcasc_repository` uncaught by `_reconcile_jcasc_config`'s narrower except clauses, crashing the reconcile hook instead of degrading to `BlockedStatus`.

src/jenkins.py
```diff
@@ -1204,3 +1206,165 @@ def sync_jcasc_config(container: ops.Container, configuration_yaml: str) -> str:
...
+        )
+        if not yaml_files:
+            raise JenkinsBootstrapError(
+                f"Path '{config_path}' not found or no YAML files in repository {url}"
+            )
+
+        merged_config: dict[str, typing.Any] = {}
+        for yaml_file in yaml_files:
+            try:
+                content = yaml.safe_load(container.pull(yaml_file, encoding="utf-8").read())
```
yhaliaw, line 1342: This `container.pull()` isn't wrapped in `except ops.pebble.PathError`, unlike every other `container.pull()` call in this file (e.g. `get_admin_credentials` line 139, `get_api_credentials` line 159, both catching `PathError`). Only `yaml.YAMLError` is caught here. If the file is removed/inaccessible between the `find` listing and this read (or the path is otherwise unreadable), `PathError` propagates uncaught out of `fetch_jcasc_repository`, crashing the reconcile hook instead of surfacing as `JenkinsBootstrapError`/`BlockedStatus`.

.github/workflows/integration_test.yaml
```diff
@@ -25,14 +25,17 @@ jobs:
       channel: 1.34-strict/stable
       extra-arguments: |
         --kube-config=${GITHUB_WORKSPACE}/kube-config
-      modules: '["test_auth_proxy.py", "test_cos.py", "test_ingress.py", "test_jenkins.py", "test_k8s_agent.py", "test_machine_agent.py", "test_plugins_part_1.py", "test_plugins_part_2.py", "test_proxy.py", "test_upgrade.py", "test_external_agent.py"]'
+      modules: '["test_jenkins.py"]'
```
yhaliaw: `modules` is reduced to just `["test_jenkins.py"]`, with the other 10 integration suites (`test_auth_proxy.py`, `test_cos.py`, `test_ingress.py`, `test_k8s_agent.py`, `test_machine_agent.py`, `test_plugins_part_1.py`, `test_plugins_part_2.py`, `test_proxy.py`, `test_upgrade.py`, `test_external_agent.py`) commented out below on line 29. If merged as-is this silently drops most of the integration coverage from CI on every future PR/scheduled run. Looks like a leftover from local debugging of this PR rather than an intended change — should this list be restored before merge?

.github/workflows/integration_test.yaml
```diff
@@ -25,14 +25,17 @@ jobs:
...
+      # "test_auth_proxy.py", "test_cos.py", "test_ingress.py", , "test_k8s_agent.py", "test_machine_agent.py", "test_plugins_part_1.py", "test_plugins_part_2.py", "test_proxy.py", "test_upgrade.py", "test_external_agent.py"]
       pre-run-script: |
         -c "sudo microk8s config > ${GITHUB_WORKSPACE}/kube-config
         chmod +x tests/integration/pre_run_script.sh
         ./tests/integration/pre_run_script.sh"
       juju-channel: 3/stable
       microk8s-addons: "dns ingress rbac hostpath-storage registry"
       with-uv: true
+      tmate-debug: true
```
yhaliaw: `tmate-debug: true` / `tmate-timeout: 90` were added here. Combined with the reduced `modules` list on line 28, this reads like leftover local-debugging config (opens a tmate SSH session on failure, adding 90 min to every CI run) rather than something intended to ship. Worth removing before merge.

src/charm.py
```diff
@@ -566,6 +568,58 @@ def _reconcile_jcasc_config(
...
+                if repo_yaml and isinstance(repo_yaml, dict):
+                    # Deep merge: for each top-level key, merge recursively
+                    for key, value in repo_yaml.items():
+                        if (
+                            key in jcasc_config
+                            and isinstance(jcasc_config[key], dict)
+                            and isinstance(value, dict)
+                        ):
+                            # Merge nested dicts
+                            jcasc_config[key].update(value)
```
yhaliaw, line 605: This is only a one-level-deep merge, not the recursive/deep merge the comment above claims. If `jcasc_config[key]` and `value` are both dicts, `.update()` overwrites any second-level key they have in common wholesale instead of merging into it — e.g. if both sources define `jenkins:` with a nested `securityRealm:` dict, the repository's `securityRealm` replaces the base config's entirely, silently dropping any sibling keys the base defined there. This happens on every reconcile, deterministically (independent of the charm object being recreated per event), and the same shallow-merge pattern is duplicated in `jenkins.py` (`fetch_jcasc_repository`, line 1357). No test exercises a same-key collision at depth ≥ 3. Worth making this an actual recursive merge, or documenting that only top-level key collisions are supported.

src/jenkins.py
```diff
@@ -1204,3 +1206,165 @@ def sync_jcasc_config(container: ops.Container, configuration_yaml: str) -> str:
...
+            except yaml.YAMLError as exc:
+                raise JenkinsBootstrapError(f"Invalid YAML in {yaml_file}: {exc}") from exc
+            if content and isinstance(content, dict):
+                for key, value in content.items():
+                    if (
+                        key in merged_config
+                        and isinstance(merged_config[key], dict)
+                        and isinstance(value, dict)
+                    ):
+                        merged_config[key].update(value)
```
yhaliaw, line 1359: Same shallow-merge issue as `charm.py` (`_reconcile_jcasc_config`, line 605): this only merges one level deep, so two YAML files that both define the same second-level nested dict under a shared top-level key (e.g. `jenkins.securityRealm`) will have the later file's sub-dict wholesale-replace the earlier one's, silently dropping sibling keys. This is deterministic per merge pass — recreation of the charm object between event hooks doesn't affect it, since `merged_config` is rebuilt from scratch on every call from the same inputs. No test covers a same-key collision at depth ≥ 3.

src/charm.py
```diff
@@ -566,6 +568,58 @@ def _reconcile_jcasc_config(
...
+                # Parse repository YAML and merge with jcasc_config
+                repo_yaml = yaml.safe_load(repo_yaml_str)
+                logger.info(
+                    "Loaded repository YAML, top-level keys: %s",
+                    list(repo_yaml.keys())
+                    if repo_yaml and isinstance(repo_yaml, dict)
+                    else "none or not a dict",
+                )
+                if repo_yaml and isinstance(repo_yaml, dict):
+                    # Deep merge: for each top-level key, merge recursively
```
m2-farzan: This comment seems wrong, deep merge implies recursive

src/jenkins.py
```diff
@@ -1204,3 +1206,167 @@ def sync_jcasc_config(container: ops.Container, configuration_yaml: str) -> str:
...
+            of YAML files.
+
+    Returns:
+        Deterministically ordered list of absolute workload YAML file paths, or
+        an empty list if the path does not exist or holds no YAML files.
+    """
+    find_cmd = ["find", config_path, "-name", "*.yaml", "-o", "-name", "*.yml"]
+    try:
+        stdout, _ = container.exec(find_cmd).wait_output()
+    except ops.pebble.ExecError:
```
m2-farzan, line 1231: Could this happen due to a permission error too?

src/jenkins.py
```diff
@@ -1204,3 +1206,167 @@ def sync_jcasc_config(container: ops.Container, configuration_yaml: str) -> str:
...
+                "Cloning JCasC repository: url=%s branch=%s config_path=%s dest=%s",
+                url,
+                branch,
+                config_path,
+                dest,
+            )
+            container.exec(clone_command, environment=environment, timeout=300).wait_output()
+            logger.info("Repository clone successful: %s", url)
+        except (ops.pebble.ChangeError, ops.pebble.ExecError) as exc:
+            # Never log exc: the command may carry the auth header. Log clean url only.
```
m2-farzan: `Never log exc` but the next line does log the exec

pebble seems to exclude the args from exception string anyways, just the command needs an update

yhaliaw (approved):

m2-farzan (commented):
Some minor comments

m2-farzan (approved):

PR #413, opened 2026-07-14
feat(haproxy): HAproxy route spoe auth

yanksyoon:
Applicable spec: <link>

### Overview

Allow Jenkins to be accessible behind HAProxy

### Rationale

To have an authentication gate

### Juju Events Changes

- HAProxy relation changed

### Module Changes

- `charm.py`: listens to and reconciles HAProxy relation

### Checklist

- [x] The [charm style guide](https://canonical.com/juju/docs/ops/latest/) was applied
- [x] The docs/changelog.md is updated with user-relevant changes in the format of [keep a changelog v1.1.0](https://keepachangelog.com/en/1.1.0/)


src/charm.py
```diff
@@ -397,6 +409,26 @@ def _reconcile_auth_proxy(self, state: State) -> None:
...
+        The requirement is published only when both the haproxy-route relation is
+        established and the external-hostname config is set. HAProxy performs
+        hostname-based routing on this value; it is also the join key for edge
+        OIDC authentication via HAProxy SPOE.
+
+        Args:
+            state: The current charm state.
+        """
+        relation = self.model.get_relation(HAPROXY_ROUTE_RELATION_NAME)
+        if relation is None or state.external_hostname is None:
```
yhaliaw: `_reconcile_haproxy_route` no-ops when `state.external_hostname is None`, but it never retracts previously-published haproxy-route data. If an operator sets `external-hostname`, gets HAProxy routing configured, then clears `external-hostname` (e.g. `juju config jenkins-k8s external-hostname=""`) while the `haproxy-route` relation stays joined, the stale hostname/service/ports stay published in the relation databag indefinitely — HAProxy keeps routing the old hostname. Nothing else in `charm.py` calls `provide_haproxy_route_requirements` to clear it, and the vendored `HaproxyRouteRequirer` only reacts to relation events, not config-changed. Worth explicitly retracting (or republishing empty) when the hostname is cleared but the relation remains.

---
*AI-assisted comment.*

src/charm.py
```diff
@@ -443,14 +475,14 @@ def _agent_discovery_url(self) -> str:
             The charm's agent discovery url.
         """
         if ingress_url := self.agent_discovery_ingress.url:
-            return ingress_url
+            return ingress_url.rstrip("/")
```
yhaliaw: `.rstrip("/")` is duplicated verbatim on both the `agent_discovery_ingress.url` branch (here) and the `server_ingress.url` fallback branch below (line 485). Since both feed the same downstream use, this could be normalized once (e.g. compute `ingress_url` first, then a single `return ingress_url.rstrip("/")`) rather than repeating it at each return site — any future ingress-shaped source added here will need to remember the same treatment otherwise the trailing-slash bug resurfaces for it.

---
*AI-assisted comment.*

src/state.py
```diff
@@ -461,6 +464,9 @@ def from_charm(cls, charm: ops.CharmBase) -> "State":
         jcasc_environment_secrets = _parse_jcasc_environment_secrets(charm)
         admin_password = _get_admin_password(charm)
 
+        external_hostname_config = typing.cast(str, charm.config.get("external-hostname", ""))
+        external_hostname = external_hostname_config.strip() or None
```
yhaliaw: This inlines the same "strip config value, empty → None" shape that this file already extracts as a named helper elsewhere, e.g. `_parse_jcasc_repository` (`repo_url = typing.cast(str, charm.config.get("jcasc-repository") or ""); return repo_url if repo_url.strip() else None`), and similarly in `_parse_jcasc_repository_token` / `_parse_jcasc_environment_secrets`. For consistency with the file's established convention, consider extracting this into a `_parse_external_hostname(charm)` helper rather than inlining it directly in `State.from_charm`.

---
*AI-assisted comment.*

tests/unit/test_haproxy_route.py
```diff
@@ -0,0 +1,147 @@
...
+
+import pytest
+from ops.testing import Harness
+
+import jenkins
+from charm import JenkinsK8sOperatorCharm
+from state import State
+
+
+def _patch_reconcile_dependencies(monkeypatch: pytest.MonkeyPatch) -> None:
```
yhaliaw: `_patch_reconcile_dependencies` is defined here but never called anywhere in this file — all four tests call `_reconcile_haproxy_route` directly instead of going through the full `_reconcile` funnel. Looks like leftover dead code from an earlier version of these tests; consider removing it (or wiring it in if there was a reason it should be exercised through the full reconcile path).

---
*AI-assisted comment.*

tests/integration/test_haproxy_route.py
```diff
@@ -0,0 +1,305 @@
...
+EXTERNAL_HOSTNAME = "jenkins.internal"
+SPOE_EXTERNAL_HOSTNAME = "jenkins-spoe.internal"
+HAPROXY_ROUTE_RELATION = "haproxy-route"
+SELF_SIGNED_CERTIFICATES_APP_NAME = "self-signed-certificates"
+
+
+@pytest_asyncio.fixture(scope="module", name="haproxy_model")
+async def haproxy_model_fixture(
+    request: pytest.FixtureRequest,
+    machine_controller: Controller,
```
yhaliaw: `haproxy_model_fixture` duplicates the existing `machine_model` fixture in `tests/integration/conftest.py` (same `add_model`/`connect`/`destroy_models(destroy_storage=True, force=True, max_wait=10*60)`/`disconnect` sequence). Since `machine_model` is `scope="module"`, this module would get its own fresh instance if it just depended on `machine_model` directly — no conflict with other test modules using the same fixture.

The two deltas here (a different model-name prefix, and honoring `request.config.option.keep_models` before destroying) aren't reasons for a separate fixture — the first is trivially parameterizable, and the second is arguably a fix that's *missing* from `machine_model_fixture` (this suite already uses `keep_models` elsewhere, e.g. `test_auth_proxy.py:104`). Folding a `prefix` param and `keep_models` support into `machine_model_fixture` and having this module depend on it directly would let both call sites share one fixture instead of carrying a near-identical copy.

---
*AI-assisted comment.*

yhaliaw (commented):
A few cleanup findings from a review pass over this diff.

---
*These review comments were generated with AI assistance.*

tests/unit/test_haproxy_route.py
yanksyoon: Removed, thanks!

tests/integration/test_haproxy_route.py
yanksyoon: Added keep_models support to shared machine_model_fixture and deleted the dupliate haproxy_model_fixture, thanks!

src/charm.py
```diff
@@ -397,6 +409,33 @@ def _reconcile_auth_proxy(self, state: State) -> None:
...
+
+        Args:
+            state: The current charm state.
+        """
+        relation = self.model.get_relation(HAPROXY_ROUTE_RELATION_NAME)
+        if relation is None:
+            return
+        if state.external_hostname is None:
+            if self.unit.is_leader():
+                relation.data[self.app].update({})
```
yhaliaw: `relation.data[self.app].update({})` with an empty dict is a no-op — it writes nothing and clears nothing from the relation databag. The docstring (lines 415–418) promises that the published haproxy-route data is retracted so HAProxy stops routing the stale hostname, but that never happens: when `external-hostname` is cleared while the relation persists, the previously published hostname, ports, and service name remain in the app databag indefinitely.

The consequence is that HAProxy continues routing the old hostname and SPOE OIDC continues using it, directly contradicting the intended behaviour. An operator who unsets `external-hostname` expecting the route to be withdrawn has no indication that it was not.

Replace the no-op with an actual clear. The charm-lib's own pattern (e.g. line 1541 of `haproxy_route.py`) is `<data_obj>.dump(relation.data[self.app], clear=True)`, but for the retraction case there is no object to dump — a direct `relation.data[self.app].clear()` is the straightforward fix:

```suggestion
                relation.data[self.app].clear()
```

*🤖 AI-assisted*

yhaliaw (commented):
Major changes needed

There are two AI finding that looks correct, can you check if they are right? Thanks.

---
🤝 Human review with AI assistance.

tests/unit/test_haproxy_route.py
```diff
@@ -0,0 +1,147 @@
...
+        harness.charm._haproxy_route, "provide_haproxy_route_requirements", provide_mock
+    )
+
+    harness.charm._reconcile_haproxy_route(State.from_charm(harness.charm))
+    provide_mock.assert_called_once()
+
+    harness.update_config({"external-hostname": ""})
+    harness.charm._reconcile_haproxy_route(State.from_charm(harness.charm))
+
+    assert harness.get_relation_data(relation_id, harness.charm.app) == {}
```
yhaliaw: The test monkeypatches `provide_haproxy_route_requirements` on line 96 — before the first `_reconcile_haproxy_route` call on line 101 — so the real write method is never invoked and the app databag is never populated. The final assertion on line 107 (`== {}`) therefore passes trivially: the databag was empty to begin with.

This means the test gives false confidence in the retraction path while simultaneously masking the real bug in `charm.py:431` (the `update({})` no-op). Even after that bug is fixed, this test cannot catch a regression because it never verifies that the databag held data in the first place.

Remove the monkeypatch for the retraction test. Let the first `_reconcile_haproxy_route` call publish real data (it will need the harness wired correctly for the lib to write), assert the databag is non-empty after the publish, then clear the hostname, reconcile again, and assert the databag is empty. That sequence will fail against the current code and will actually validate the retraction once the fix is in place.

*🤖 AI-assisted*

yhaliaw (approved):

